Issue #2, opened 2019-05-29
Define the interfaces for all the functions in Config


deepesh-kn:
Config provides all the data that is needed to run the facilitator. It reads data from `mosaic-config` and `facilitator-config` JSON files and provides getter functions that can be used in other parts of the facilitator code base.

# **This ticket is about defining the interfaces of all the functions in the Config.**

The proposed functions are provided for reference. 
The implementer can think this in detail and can add/remove functions.

- [ ] **constructor**
	 - Accepts `mosaic-config` and `facilitator-config` as params
	 - Validates `config` (against schema for `facilitator-config : mosaic-config`)
	 - Validate worker addresses
<br/>
<br/>

- [ ] **function to get DB path.**
[Check DB tickets if more functions related to DB are needed]
<br/>

- [ ] **function to get RPC endpoints for the given chain.**
<br/>

- [ ] **function to get all the worker addresses for the given chain.**
<br/>

- [ ] **function to get `encrypted_account` for a given address.**
<br/>

- [ ] **function to get OSTComposer contract address**
<br/>

- [ ] **function to get all Gateway contract addresses.**
<br/>

- [ ] **function to get Gateway contract address for a given chain.**
<br/>

- [ ] **function to get CoGateway contract address for a given chain.**
<br/>

- [ ] **function to get worker addresses for a given chain.**
<br/>

________________
**Example of mosaic-config.json is given below.**

```
{
  "originChain": {
    "chainId": 3,
    "contractAddresses": { // Contracts deployed on origin chain, should be reused for all auxiliary chains. 
      "simpleTokenAddress": "",
      "merklePatriciaLibAddress": "",
      "GatewayLibAddress": "",
      "messageBusAddress": ""
    }
  },
  "auxiliaryChains": {
    "1406": {
      "chainId": 1406,
      "bootNodes": [], //List of boot nodes
      "genesis": {}, 
      "contractAddresses": { // Contract addresses specific to single auxiliary chain 
        "origin": { // Contracts deployed on origin chain specific to single auxiliary chain
          "anchorOrganizationAddress": "",
          "anchorAddress": "",
          "ostGatewayOrganizationAddress": "",
          "ostEIP20GatewayAddress": "",
          "ostComposerAddress": "" 
        },
        "auxiliary": { // Contracts deployed on the auxiliary chain
          "ostPrimeAddress": "",
          "anchorOrganizationAddress": "",
          "anchorAddress": "",
          "merklePatriciaLibAddress": "",
[...]

PR #63, opened 2019-06-12
Config interface

deepesh-kn:
This PR defines the interfaces for Config class.
- Dummy MosaicConfig.ts is added in this repo. This will be replaced by MosaicConfig class from mosaic-chains.

fixes: #2


abhayks1:
Reviewing...

src/Account.ts
```diff
@@ -0,0 +1,25 @@
...
+     */
+    constructor(
+        readonly address: string,
+        readonly encryptedKeyStore: EncryptedKeystoreV3Json
+    ) { }
+
+    /**
+     * Unlocks this account and keeps it in memory unlocked.
+     * @param web3 The web3 instance that this account uses.
+     * @param password The password required to unlock the account.
```
abhayks1: return type boolean is not specified in the documentation.

src/Config.ts
```diff
@@ -0,0 +1,119 @@
...
+            }
+        }
+        return null;
+    }
+}
+
+/**
+ * Holds database configurations.
+ */
+export class Database {
```
abhayks1: I feel we should call the class DBConfig as it only holds the config to make a db connection.

src/Config.ts
```diff
@@ -0,0 +1,119 @@
...
+    /** Database password */
+    private _password: string;
+
+    /**
+     * Constructor.
+     * @param facilitatorConfig Facilitator config json data.
+     */
+    public constructor(facilitatorConfig: Record<string, any>) {
+        let dbJson:Record<string, any> = facilitatorConfig.database;
+        this.path = dbJson.path;
```
abhayks1: we should add dbType attribute here also.

src/Config.ts
```diff
@@ -0,0 +1,119 @@
...
+
+    /** Database password */
+    private _password: string;
+
+    /**
+     * Constructor.
+     * @param facilitatorConfig Facilitator config json data.
+     */
+    public constructor(facilitatorConfig: Record<string, any>) {
+        let dbJson:Record<string, any> = facilitatorConfig.database;
```
abhayks1:
```suggestion
        const dbJson:Record<string, any> = facilitatorConfig.database;
```

src/Config.ts
```diff
@@ -0,0 +1,119 @@
...
+    /** Worker account object. */
+    public worker: Account;
+
+    /**
+     * Constructor.
+     * @param chainId Chain id.
+     * @param facilitatorConfig Facilitator config json data.
+     */
+    public constructor(readonly chainId: string, facilitatorConfig: Record<string, any>) {
+        let chainData = facilitatorConfig.chains[chainId];
```
abhayks1:
```suggestion
        const chainData = facilitatorConfig.chains[chainId];
```

src/Config.ts
```diff
@@ -0,0 +1,119 @@
...
+    }
+}
+
+/**
+ * Holds database configurations.
+ */
+export class Database {
+
+    /** Database path */
+    public path: string;
```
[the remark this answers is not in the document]
deepesh-kn: Added host also in the config.

src/Config.ts
```diff
@@ -0,0 +1,119 @@
...
+
+    /** Database user name */
+    public userName: string;
+
+    /** Database password */
+    private _password: string;
```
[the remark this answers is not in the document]
deepesh-kn: I have added a getter for `password`, so the internal variable is `_password`

src/Config.ts
```diff
@@ -0,0 +1,135 @@
...
+  public database: DBConfig;
+
+  public chains: Record<string, any>;
+
+  public mosaic: MosaicConfig;
+
+  /**
+   * Constructor.
+   * @param mosaicConfigPath Mosaic config path.
+   * @param facilitatorConfigPath Facilitator config path.
```
benjaminbollen: is it not a facilitator manifest?

src/Config.ts
```diff
@@ -0,0 +1,135 @@
...
+    }
+    return chains;
+  }
+
+  /**
+   * Get config json data from the given file path.
+   * @param filePath Facilitator config file path.
+   * @returns JSON data for facilitator config.
+   */
+  private getConfigJsonFromPath(filePath: string): Record<string, any> {
```
benjaminbollen: `getManifestFromPath` and should be in yaml?

src/Config.ts
```diff
@@ -0,0 +1,135 @@
...
+    this.host = dbJson.host;
+    this.userName = dbJson.user_name;
+    this._password = dbJson.password;
+  }
+
+  /**
+   * Get the password for the database.
+   */
+  get password(): string {
+    return process.env.MOSAIC_FACILITATOR_DB_PASSWORD || this._password;
```
benjaminbollen: you defined a const for this https://github.com/OpenST/facilitator/pull/63/files#diff-5eb68ff2f89ff29d7d6ff9be9116e1beR6?

src/Config.ts
```diff
@@ -0,0 +1,135 @@
...
+  public mosaic: MosaicConfig;
+
+  /**
+   * Constructor.
+   * @param mosaicConfigPath Mosaic config path.
+   * @param facilitatorConfigPath Facilitator config path.
+   */
+  public constructor(
+    mosaicConfigPath: string,
+    facilitatorConfigPath: string,
```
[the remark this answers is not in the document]
deepesh-kn: As per our developer guidelines, we are using `airbnb` styles, it specifies to have a trailing comma.

src/Config.ts
```diff
@@ -0,0 +1,135 @@
...
+  ) {
+    this.mosaic = MosaicConfig.fromPath(mosaicConfigPath);
```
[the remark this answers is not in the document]
deepesh-kn: There is an onother ticket that will validate the config and It will throw the error.

src/Config.ts
deepesh-kn: User may not add the accounts and encrypted keys manually. They are generated by the code.
The encrypted account will be JSON, so it will become a mix of YAML and JSON as a string.

I still think that its config.
The file defined the configuration to run the facilitator.

src/Config.ts
deepesh-kn: Good catch 👍 , changes done

src/Config.ts
deepesh-kn: Manifest to me sounds like it should contain the version, app name, package identifier. etc.
I still think that its config.
The file defines the configuration to run the facilitator.

I may be wrong here though.

src/Config.ts
benjaminbollen: Ok, makes sense

Issue #53, opened 2019-06-07
Implement stake requests model class


deepesh-kn:
Implement stake requests model class

- stake_request_hash
- message_hash
- amount
- beneficiary
- gasPrice
- gasLimit
- nonce
- gateway
- staker_proxy (-> sender in gateway, for messagehash)

Abstract the database interaction layer.
(StakeRequestRepository, Query… etc remember the discussion)

pgev:
In a draft PR (#66) DatabaseWrapper and StakeRequestRepository main classes are created. 

DatabaseWrapper gives a promisified interface to sqlite3.Database object to ease usage of database operations.
StakeRequestRepository enables to create a stake_requests table in the database, to create a new entry (stake request) and retrieve stake request by stake request' hash.

StakeRequestRepository class is covered by unit tests.
DatabaseWrapper class does not have unit tests, however, functions are mostly tested through StakeRequestRepository unit test. We should think what kind of tests do we want for DatabaseWrapper class.

DatabaseWrapper is documented.
StakeRequestRepository class is missing documentation.

PR #66, opened 2019-06-14
StakeRequest model (with a direct usage of sqlite)

pgev:
Implements DatabaseWrapper and StakeRequestRepository classes.
DatabaseWrapper abstracts away and promisifies database operations.
StakeRequestRepository is a model to work with stake_request database table.

Fixes: #53


deepesh-kn:
I am in favor of using ORM. `sequelize` looks a good option 👍

pgev:
> I am in favor of using ORM. `sequelize` looks a good option 👍

@deepesh-kn could you please elaborate in which way. ORM as an interface or as an implementation? @sarvesh-ost was thinking (Sarvesh, please correct me here), to use sequelize for implementation to ease manipulations with database. However, ORM is much broader concept, and it's also possible to use it as an interfaces in the project.

src/data_access_objects/StakeRequestRepository.ts
```diff
@@ -0,0 +1,128 @@
...
+      + `${StakeRequestRepository.nonceColumnName}, `
+      + `${StakeRequestRepository.gatewayColumnName}, `
+      + `${StakeRequestRepository.stakerProxyColumnName} `
+      + ') '
+      + 'VALUES '
+      + '( '
+      + `'${stakeRequest.stakeRequestHash}', `
+      + `'${stakeRequest.messageHash}', `
+      + `${stakeRequest.amount}, `
+      + `'${stakeRequest.beneficiary}', `
```
benjaminbollen: extra single quotes?

benjaminbollen: nope, different types :D

PR #68, opened 2019-06-17
Transaction handler

### LINKED ISSUE #56: Implement transaction handler
Implement TransactionHandler

- Receives the call back from graph-node.
- Creates(inserts)/updates the model object. (Implementer can think of this)
- Calls the service class to perform actions. While developing this, we can just have a no-op.
- The events that it will receive are
        - StakeRequested
        - StakeRequestAccepted
        - StakeRequestRejected
        - StakeRequestRevoked
        - StateRootAnchored
        - StakeIntentDeclared
        - StakeIntentConfirmed
        - ProgressStake
        - ProgressMint
        - GatewayProven

sarvesh-ost:
Implementation of transaction handler. 

Defines a transaction handler which can handle different kinds of transaction. 
This PR contains an example handler i.e. StakeRequestHandler to demonstrate how to write handlers. 
Fixes #56


puneet-khushwani-eth:
Stating to review this.

src/handlers/HandlerFactory.ts
```diff
@@ -0,0 +1,6 @@
+import StakeRequestedHandler from './StakeRequestedHandler';
```
puneet-khushwani-eth: As  TransactionHandler.ts takes `handlers` as Constructor args, not sure how do we plan to use  this class. Can you please elaborate how do we plan to use it.

src/models/StakeRequest.ts
```diff
@@ -0,0 +1,79 @@
+import { Address, Bytes32 } from '../types';
+
+export default class StakeRequest {
+  private _id: string;
```
puneet-khushwani-eth: By not adding a `readonly` here, are we leaving the possibility to modify the states via some other functions ? 

Ex. from the event we get via TransactionHandlers, we create an instance. We then from our services modify the states of the same instance and pass it over to Repository classes to sync this state in Database ?

src/types.ts
```diff
@@ -0,0 +1,2 @@
+export type Bytes32 = string;
+export type Address = string;
```
puneet-khushwani-eth:
```suggestion
export type Address = string;

```

src/models/StakeRequest.ts
```diff
@@ -0,0 +1,79 @@
...
+
+  private _amount: string;
+
+  private _beneficiary: Address;
+
+  private _gasPrice: string;
+
+  private _gasLimit: string;
+
+  private _nonce: string;
```
puneet-khushwani-eth: Curious to know why did we append _ to the names of all private variables ?

src/handlers/HandlerFactory.ts
sarvesh-ost: TransactionHandler needs different kinds of handlers may be 10 for now. It seems it's not the responsibility of transaction handler to instantiate different handler. I created this factory class which can instantiate different kinds of handlers and which is passed in the constructor.

src/models/StakeRequest.ts
sarvesh-ost: variables are private so that can't be accessed outside the class.

src/models/StakeRequest.ts
sarvesh-ost: TS compiler will give an error because of duplicate declaration one with getter and other with the variable declaration. Adding `_` in the parameter to variable resolves it.
```
public get id(): string {
    return this._id;
  }
```
and 
```
private id: string;
```

src/handlers/HandlerFactory.ts
puneet-khushwani-eth: Agree this approach works well with TDD.

pgev:
Reviewing.

abhayks1:
Reviewing 🐮 🐃

src/TransactionHandler.ts
```diff
@@ -0,0 +1,28 @@
...
+  public constructor(handlers: Record<string, ContractEntityHandler<any>>) {
+    this.handlers = handlers;
+    this.handle = this.handle.bind(this);
+  }
+
+  /**
+   * This method accept bulkTransactions and handles them with specific handlers.
+   * @param bulkTransactions List of bulkTransactions.
+   */
+  public handle(bulkTransactions: any): void {
```
abhayks1: Since subscription code is merged, the method can be integrated with GraphClient.subscribe method 🏗 🏛

tests/TransactionHandler/handle.test.ts
```diff
@@ -0,0 +1,46 @@
+import { assert } from 'chai';
+import TransactionHandler from '../../src/TransactionHandler';
+import StakeRequestedHandler from '../../src/handlers/StakeRequestedHandler';
+import SpyAssert from '../SpyAssert';
+
+import sinon = require('sinon');
```
abhayks1: This gives linting error with typescript 🔴 . 

Below works correctly 🍏 
import * as sinon from 'sinon';

src/handlers/ContractEntityHandler.ts
```diff
@@ -0,0 +1,26 @@
...
+  /**
+   * This method parse a transaction and returns a model
+   * @param any Transaction from the subscriber.
+   */
+  abstract parse(any: any[]): T[];
+
+  /**
+   * This is a template method, this method shouldn't be override by implementer.
+   * @param any Transaction from the subscriber.
+   */
```
abhayks1:
```suggestion
  @return {void}
   */
```

src/handlers/ContractEntityHandler.ts
```diff
@@ -0,0 +1,26 @@
...
+  abstract handle(models: T[]): void;
+
+  /**
+   * This method parse a transaction and returns a model
+   * @param any Transaction from the subscriber.
+   */
+  abstract parse(any: any[]): T[];
+
+  /**
+   * This is a template method, this method shouldn't be override by implementer.
```
abhayks1:
```suggestion
   * This is a template method, this method shouldn't be overridden by implementer.
```

src/handlers/HandlerFactory.ts
```diff
@@ -0,0 +1,6 @@
...
+
+
+export default {
+  stakeRequesteds: new StakeRequestedHandler(),
```
abhayks1: This can be merged with Facilitator.getSubscriptionDetails() method.
We have similar information there. We can merge below to a single class/method.
So for stakeRequested entity => Subscription query => StakeRequestedHandler.
Object can look like below:
```
{
  stakeRequested: { subscriptionQuery: "", handler: StakeRequestedHandler}
}
```
You can decide on the right placeholder 🚮

src/handlers/StakeRequestedHandler.ts
```diff
@@ -0,0 +1,43 @@
...
+      nonce,
+      staker,
+      gateway,
+      stakeRequestHash,
+    );
+  });
+
+  /**
+   * This method defines action on receiving stake request model.
+   * @param stakeRequest instance of StakeRequest model .
```
abhayks1:
```suggestion
   * @param stakeRequest instance of StakeRequest model.
```

src/handlers/StakeRequestedHandler.ts
```diff
@@ -0,0 +1,43 @@
...
+   */
```
abhayks1:
```suggestion
    @return <void>
   */
```

src/handlers/StakeRequestedHandler.ts
```diff
@@ -0,0 +1,43 @@
...
+  public handle = (stakeRequest: StakeRequest[]): void => {
+
+    // stakeRequestRepository.save(stakeRequest);
```
abhayks1: There is no save method in StakeRequestRepository. There is create method. Please see the updated StakeRequestRepository.ts code.

src/models/StakeRequest.ts
```diff
@@ -0,0 +1,79 @@
+import { Address, Bytes32 } from '../types';
+
+export default class StakeRequest {
+  private _id: string;
+
+  private _amount: string;
```
abhayks1: Please see the updated code. These properties are already defined in model/repository classes.
We might not need this class.

src/handlers/HandlerFactory.ts
sarvesh-ost: If you see latest implementation of HandlerFactory, it's way more complex. This now has responsibility of instantiating all the handler which would require instantiation of repositories and services.

src/handlers/StakeRequestedHandler.ts
sarvesh-ost: It was an example comment 🤦‍♂

src/GraphClient.ts
```diff
@@ -4,10 +4,11 @@ import { InMemoryCache, NormalizedCacheObject } from 'apollo-cache-inmemory';
 import { SubscriptionClient } from 'subscriptions-transport-ws';
 import { Subscription } from 'apollo-client/util/Observable';
 import gql from 'graphql-tag';
-import * as WebSocket from 'ws'
+import * as WebSocket from 'ws';
 
 import Logger from './Logger';
-//import WebSocket = require('ws');
+import TransactionHandler from './TransactionHandler';
+// import WebSocket = require('ws');
```
deepesh-kn: Please remove this.

src/models/StakeRequestRepository.ts
```diff
@@ -159,6 +159,28 @@ export class StakeRequestRepository {
...
+   * @param stakeRequestsAttributes Array of stake request attributes.
+   *
+   * @return Array of stake request object.
+   */
+  public async bulkCreate(stakeRequestsAttributes:
+  StakeRequestAttributes[]): Promise<StakeRequest []> {
+    try {
+      const stakeRequests: StakeRequest [] = await StakeRequestModel.bulkCreate(
+        stakeRequestsAttributes,
+        { ignoreDuplicates: true },
```
deepesh-kn: How to handle the duplicates?
 I think that this should be like an `insert or update`, instead of just `insert`.

test/TransactionHandler/handle.test.ts
```diff
@@ -0,0 +1,65 @@
...
+        __typename: 'StakeRequested',
+        amount: '0',
+        beneficiary: '0x79376dc1925ba1e0276473244802287394216a39',
+        gasLimit: '2',
+        gasPrice: '1',
+        gateway: '0x4e4ea3140f3d4a07e2f054cbabfd1f8038b3b4b0',
+        id: '0xa80c3db5089412e553b3b4defc3b3759f56b3a77257be6940251a7a05b5c4fec-0',
+        nonce: '1',
+        stakeRequestHash: '0xdc67e167a7dd111e4f2c27796ceb89955bb68b995eef3a84aa86b38a5f7cd22c',
+        staker: '0x79376dc1925ba1e0276473244802287394216a39',
```
deepesh-kn: Addresses and hashes are actual data. Let us use data that can be identified as the dummy
```
{
        __typename: 'StakeRequested',
        amount: '0',
        beneficiary: '0x0000000000000000000000000000000000000001',
        gasLimit: '2',
        gasPrice: '1',
        gateway: '0x0000000000000000000000000000000000000002',
        id: '0x0000000000000000000000000000000000000000000000000000000000000002-0',
        nonce: '1',
        stakeRequestHash:'0x0000000000000000000000000000000000000000000000000000000000000001',
        staker: '0x0000000000000000000000000000000000000003',
}
```

test/handlers/StakeRequestedHandler/persist.test.ts
```diff
@@ -0,0 +1,51 @@
...
+      amount: '10',
+      beneficiary: '0x7f14BcAFdF55a45Fd64384e3496b62Ca8A1B099D',
+      gasPrice: '1',
+      gasLimit: '1',
+      nonce: '1',
+      staker: '0xd328C7bE0D524aa33b8118163C431dCd0d46EE82',
+      gateway: '0xF1e701FbE4288a38FfFEa3084C826B810c5d5294',
+      stakerProxy: '0xE1e701FbE4288a38FfFEa3084C826B810c5d5294',
+      stakeRequestHash: Utils.sha3('1'),
+    }];
```
deepesh-kn:
```
const transactions = [{
      id: '1',
      amount: '10',
      beneficiary: '0x0000000000000000000000000000000000000001',
      gasPrice: '1',
      gasLimit: '1',
      nonce: '1',
      staker: '0x0000000000000000000000000000000000000002',
      gateway: '0x0000000000000000000000000000000000000003',
      stakerProxy: '0x0000000000000000000000000000000000000004',
      stakeRequestHash: Utils.sha3('1'),
    }];
```

src/handlers/StakeRequestedHandler.ts
```diff
@@ -0,0 +1,65 @@
...
+  public constructor(stakeRequestRepository: StakeRequestRepository) {
+    super();
+    this.stakeRequestRepository = stakeRequestRepository;
+    this.persist = this.persist.bind(this);
+  }
+
+  /**
+   * This method parse stakeRequest transaction and returns stakeRequest model object.
+   *
+   * @param transactions Transaction objects.
```
abhayks1:
```suggestion
   * @param {any} transactions Transaction objects.
```

src/handlers/StakeRequestedHandler.ts
```diff
@@ -0,0 +1,65 @@
...
+      };
+    });
+    this.stakeRequestRepository.bulkCreate(models);
+    return models;
+  };
+
+  /**
+   * This method defines action on receiving stake request model.
+   *
+   * @param stakeRequest array of instances of StakeRequestAttributes object.
```
abhayks1:
```suggestion
   * @param {StakeRequestAttributes[]} stakeRequest array of instances of StakeRequestAttributes object.
```

src/GraphClient.ts
```diff
@@ -31,9 +32,10 @@ export default class GraphClient {
    * Documentation: https://www.apollographql.com/docs/react/advanced/subscriptions/
    *
    * @param {string} subscriptionQry Subscription query.
+   * @param {TransactionHandler} handler Transaction handler object.
    * @return {Subscription} Query subscription object.
    */
-  public subscribe(subscriptionQry: string): Subscription {
+  public subscribe(subscriptionQry: string, handler: TransactionHandler): Subscription {
```
abhayks1: TransactionHandler is never going to change. It's always a singleton object. Shouldn't this be part of constructor.
While creating instance of GraphClient in facilitator start we can pass the handler object in the constructor. 
What do you think?

src/TransactionHandler.ts
```diff
@@ -0,0 +1,40 @@
...
+    });
+
+    await Promise.all(persistPromises);
+
+    Object.keys(models).forEach((transactionKind) => {
+      const handler = this.handlers[transactionKind];
+      handler.handle(models[transactionKind]);
+    });
+  }
+}
```
abhayks1: There are linting errors in TransactionHandler, HandlerFactory, bulkcreate.test.ts, StubData.ts

src/models/StakeRequestRepository.ts
sarvesh-ost: It will ignore duplicate records based on primary key. Only insert non duplicate.

test/TransactionHandler/handle.test.ts
sarvesh-ost: why?

src/models/StakeRequestRepository.ts
deepesh-kn: So if any other fields other than primary key in the record are different it will not update/insert the record?
I think that we need to create a nonexisting record and update an existing record.

test/TransactionHandler/handle.test.ts
deepesh-kn: This is dummy data and not actual data.

src/handlers/HandlerFactory.ts
```diff
@@ -0,0 +1,19 @@
+import StakeRequestedHandler from './StakeRequestedHandler';
+import Database from '../models/Database';
+
+export default class HandlerFactory {
+  /**
+   * This methods instantiate different kinds of handlers. This method also takes
+   * responsibility of instantiating repositories and services.
+   *
+   * @param db Database connection.
```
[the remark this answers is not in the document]
sarvesh-ost: This is not needed, as typescript defines types.

src/GraphClient.ts
sarvesh-ost: This way also, it's a singleton. I think handler should be passed along with the query.

src/handlers/HandlerFactory.ts
```diff
@@ -0,0 +1,19 @@
...
+   * @return Different kinds of transaction handlers.
+   */
+  public static get(db: Database): {stakeRequesteds: StakeRequestedHandler} {
```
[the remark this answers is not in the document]
sarvesh-ost: It's not huge now, can be done when it will be huge.

src/GraphClient.ts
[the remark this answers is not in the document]
sarvesh-ost: Not sure on this.  These kind of validation makes sense when dependency is passed from outside by user.

test/TransactionHandler/handle.test.ts
sarvesh-ost: This is also dummy, why do you think it's actual.

src/models/StakeRequestRepository.ts
sarvesh-ost: I am not sure of this. can you give example of use case?

src/handlers/HandlerFactory.ts
abhayks1: I still think we should merge entity wise handler, subscription queries together  to have single placeholder. Easy to access and read
There will be some refactoring needed in Facilitator.getSubscriptionDetails. We can take up at that time.

src/TransactionHandler.ts
```diff
@@ -0,0 +1,40 @@
+import ContractEntityHandler from './handlers/ContractEntityHandler';
+import { HandlerNotFoundException } from './Exception';
+
+export default class TransactionHandler {
+  private readonly handlers: Record<string, ContractEntityHandler<any>>;
+
+  public constructor(handlers: Record<string, ContractEntityHandler<any>>) {
```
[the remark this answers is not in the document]
abhayks1: Documentation missing for argument ContractEntityHandler ⏰

src/handlers/HandlerFactory.ts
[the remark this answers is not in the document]
abhayks1: This means we don't need to define datatype for function arguments and return type as a practice in our codebase. Right?
cc: @deepesh-kn

test/models/StakeRequestRepository/bulkcreate.test.ts
```diff
@@ -0,0 +1,115 @@
...
+    );
+
+    assert.strictEqual(
+      stakeRequestResponse.length,
+      4,
+      'Only one record should be returned from database',
+    );
+  });
+});
+
```
abhayks1: There are 2 new lines in the end 👁

src/handlers/HandlerFactory.ts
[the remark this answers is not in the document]
sarvesh-ost: Yes 👍

test/TransactionHandler/handle.test.ts
deepesh-kn: By seeing I can identify that its a dummy data. 
Otherwise, I will expect that there is some significance of particular value used and why that value was selected. We had this discussion in the past. 

In this example, if we are going to validate the stake request hash then it will not be a dummy value.
Anyways it's not that critical. Will add it to developer guidelines.

src/models/StakeRequestRepository.ts
deepesh-kn: As discussed the entity handler will take care of what is to be updated and inserted.

src/handlers/HandlerFactory.ts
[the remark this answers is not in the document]
abhayks1: We should update this is typescript dev guideline.

PR #69, opened 2019-06-17
Subscription and unsubscription to deployed subgraphs

### LINKED ISSUE #50: Subscribe and unsubscribe to graph node instance.
Subscribe/Unsubscribe to all entities with one subscriber per graph-node instance (ie. per chain)

abhayks1:
Discussed with thegraph team on Discord and they have recommended below client:
https://www.apollographql.com/docs/react/advanced/subscriptions/

With the subscription it returns all data matching the query for every new entity(stakeRequested..) event. Ideally it should return only the incremental data. e.g. Should return only new stakeRequested event data.

Created hackmd doc with troubleshooting examples and learnings while integrating the client. 
**Link**: https://hackmd.io/mVF9i4B7S3O2Iq-Z_leFMg

Please see more details on the ticket #50 

Resolves #50


abhayks1:
Looking into the travis issue ⏰

src/Facilitator.ts
```diff
@@ -17,20 +24,43 @@ export default class Facilitator {
...
+    }
+  }
 
+  /**
+   * List of all queries to subscribe.
+   *
+   * @return {string[]} List of subscription queries
+   */
+  private getSubscriptionQueries() {
+    return ['subscription{stakeRequesteds{id}}'];
```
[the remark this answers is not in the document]
abhayks1: We will need to update the schema in mosaic chains. So wherever StakeRequested is there needs to be updated with requestedStake.

abhayks1:
Travis issue fixed 🍏

src/Facilitator.ts
```diff
@@ -17,20 +21,60 @@ export default class Facilitator {
...
+   * Populate subscription queries
+   *
+   * @return {string[]} List of subscription queries.
+   */
+  private getSubGraphDetails() {
+    return {
+      origin: {
+        subGraphEndPoint: 'http://localhost:8000/subgraphs/name/openst/ost-composer',
+        subscriptionQueries: ['subscription{stakeRequesteds{id amount beneficiary gasLimit' +
+        ' gasPrice gateway nonce staker stakeRequestHash }}'],
```
deepesh-kn: I think that we should have subscriptionQueries as key-value pair instead of an array.
This will provide the subscriber flexibility to select particular subscription query.

```
subscriptionQueries: {
    'stakeRequested': 'subscription{stakeRequesteds{id amount beneficiary gasLimit' +
        ' gasPrice gateway nonce staker stakeRequestHash }}',
    'xyz': 'pqr'
}
```

src/Facilitator.ts
```diff
@@ -17,20 +21,60 @@ export default class Facilitator {
...
+    }
+  }
 
+  /**
+   * List of all queries to subscribe.
+   * Add GQL queries needed for the model/services.
+   * Replace it with subgraph endpoint from config after it's populated.
+   * Populate subscription queries
+   *
+   * @return {string[]} List of subscription queries.
```
deepesh-kn: The documentation and return type appear to be incorrect.

src/GraphClient.ts
```diff
@@ -0,0 +1,75 @@
...
+
+  /**
+   * Subscribes to the input subscription subgraph query and delegates the response
+   * to observer i.e. TransactionHandler.
+   * Documentation: https://www.apollographql.com/docs/react/advanced/subscriptions/
+   *
+   * @param subscriptionQry Subscription query object.
+   * @return `Subscription` Query subscription object
+   */
+  public subscribe(subscriptionQry: string) {
```
deepesh-kn: Specify return type

src/GraphClient.ts
```diff
@@ -0,0 +1,75 @@
...
+    if( !subscriptionQry ) {
+      const err = new TypeError("Mandatory Parameter 'subscriptionQry' is missing or invalid.");
+      return Promise.reject(err);
```
deepesh-kn: Why not throw error?
```suggestion
      throw err;
```

src/GraphClient.ts
```diff
@@ -0,0 +1,75 @@
...
+        reconnect: true,
+      },
+      WebSocket
+    );
+    // Creates WebSocket link.
+    const wsLink = new WebSocketLink(subscriptionClient);
+    // Instantiate in memory cache object.
+    const cache = new InMemoryCache();
+    // Instantiate apollo client
+    return new ApolloClient({ link: wsLink, cache });
```
deepesh-kn: Can a single ApolloClient object subscribe to multiple queries?
If yes then we don't need to create a new object every time.

test/Facilitator/constructor.test.ts
```diff
@@ -0,0 +1,20 @@
...
+import Facilitator from './../../src/Facilitator'
+
+describe('Facilitator.constructor()', () => {
+
+  it('should construct with correct parameters', async () => {
+    const configStub = sinon.createStubInstance(Config);
+    const dbConnection = sinon.spy();
+    const facilitator = new Facilitator(configStub, dbConnection);
+
+    assert(facilitator);
```
deepesh-kn: Check the values that are set in the constructor.

test/Facilitator/start.test.ts
```diff
@@ -0,0 +1,31 @@
...
+
+import { assert } from  'chai'
+const sinon = require('sinon');
+
+import GraphClient from './../../src/GraphClient';
+import { Config } from './../../src/Config'
+import Facilitator from './../../src/Facilitator'
+import SpyAssert from './../utils/SpyAssert'
+
+describe('Facilitator.stop()', () => {
```
deepesh-kn:
```suggestion
describe('Facilitator.start()', () => {
```

test/Facilitator/start.test.ts
```diff
@@ -0,0 +1,31 @@
...
+      graphClient,
+      'subscribe',
+      sinon.fake.resolves(true),
+    );
+
+    const configStub = sinon.createStubInstance(Config);
+    const dbConnection = sinon.spy();
+    facilitator = new Facilitator(configStub, dbConnection);
+
+    await facilitator.start();
```
deepesh-kn: Asserts are missing? 🤔

test/Facilitator/stop.test.ts
```diff
@@ -0,0 +1,21 @@
...
+import Facilitator from './../../src/Facilitator'
+
+describe('Facilitator.stop()', () => {
+  let facilitator;
+
+  it('should stop facilitator gracefully', async () => {
+    const configStub = sinon.createStubInstance(Config);
+    const dbConnection = sinon.spy();
+    facilitator = new Facilitator(configStub, dbConnection);
+    await facilitator.stop();
```
deepesh-kn: Asserts are missing.

test/GraphClient/constructor.test.ts
```diff
@@ -0,0 +1,18 @@
...
+describe('GraphClient.constructor()', () => {
+  let subgraphEndPoint;
+
+  beforeEach(() => {
+    subgraphEndPoint = '/subgraph/v1/subgraph1';
+  });
+
+  it('should construct with correct parameters', async () => {
+    const graphClient = new GraphClient(subgraphEndPoint);
+    assert(graphClient);
```
deepesh-kn: Assert `subgraphEndPoint` is set or not.

src/GraphClient.ts
```diff
@@ -0,0 +1,75 @@
...
+import { SubscriptionClient } from 'subscriptions-transport-ws';
+import gql from 'graphql-tag';
+
+import WebSocket = require('ws');
+
+/**
+ * The class interacts with graph node server for subscription and query.
+ */
+export default class GraphClient {
+  private subgraphEndPoint: string;
```
deepesh-kn: Why is this private, this can be just read-only property.

test/GraphClient/subscribe.test.ts
```diff
@@ -0,0 +1,43 @@
...
+      graphClient,
+      'getClient',
+      sinon.fake.resolves(true),
+    );
+    const spySubscribe = sinon.replace(
+      graphClient,
+      'subscribe',
+      sinon.fake.resolves(true),
+    );
+    const querySubscriber = await graphClient.subscribe(subscriptionQry);
```
deepesh-kn: in line 27 the `subscribe` function is already mocked. So this test is incorrect.

src/GraphClient.ts
abhayks1: Subscription is the return type already specified. Am I missing something here 😕 ?

src/GraphClient.ts
abhayks1: AssertAsync works with Promise.reject else we will need to write custom callback to handle throw.  We have been using this in other repos also.

src/GraphClient.ts
abhayks1: Facilitator.start() expects originSubGraphClient and auxiliarySubGraphClient for each chain sub graph endpoint. Each subgraph client expects single apollo client in the constructor. So apollo client will be created per subgraph 💭 .

src/GraphClient.ts
abhayks1: In the updated code apolloClient is the private variable which is not needed outside as per current use cases.

test/GraphClient/constructor.test.ts
abhayks1: subgraphEndPoint is not a property of GraphClient in the updated code.

src/Facilitator.ts
abhayks1: makes sense 👍

test/Facilitator/start.test.ts
abhayks1: test for start/stop is not needed because there is nothing much start is doing after the logic has been moved to Subscriber class.

test/GraphClient/subscribe.test.ts
abhayks1: It's taken care 🚀

src/Facilitator.ts
```diff
@@ -20,17 +26,59 @@ export default class Facilitator {
...
+  }
 
+  /**
+   * Subgraph details object which contains chain based subGraphEndPoitn & subscriptionQueries.
+   * Note: Replace subGraphEndPoint from Config.ts. It should come from Config:Chain class.
+   * Feel free to add subscription queries.
+   *
+   * @return <any> Object containing chain based subscriptionQueries.
+   */
+  public static getSubscriptionDetails() {
```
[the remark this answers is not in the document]
abhayks1: Makes sense 👍. However I think the current values are hardcoded and we are thinking to populate it from different sources. The function might also be not needed based on the discussion. We can take this up at that time.

src/GraphClient.ts
```diff
@@ -0,0 +1,79 @@
...
+    return querySubscriber;
+  }
+
+  /**
+   * Creates and returns graph client.
+   *
+   * @param {string} subgraphEndPoint Subgraph endpoint.
+   * @return {ApolloClient<NormalizedCacheObject>}
+   */
+  public static getClient(subgraphEndPoint: string) {
```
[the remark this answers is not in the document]
abhayks1: Sounds a good optimization solution 👍. I don't think  we need this feature now 🕐as we are creating only two graph client objects  per subgraph for a facilitator..

src/Subscriber.ts
```diff
@@ -0,0 +1,52 @@
+import { Subscription } from 'apollo-client/util/Observable';
+import GraphClient from './GraphClient';
+
+/**
+ * Subscriber class subscribes and unsubscribes subscription queries of a subgraph.
+ */
+export default class Subscriber {
+  public querySubscriptions: Subscription[];
+
+  private subscriptionQueries: string[];
```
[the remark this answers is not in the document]
abhayks1: Very good  point 💯. But I don't think we have currently use case of subscribe and unsubscribe from multiple sources. Only facilitator calls subscribe on start and unsubscribe is called on SIGINT. 
If you feel strongly 💪  that this use case can come in near future I can make the change.

abhayks1:
Available for review again 🚀 🐶

src/GraphClient.ts
deepesh-kn: Yes, its missing in the following line.
`public subscribe(subscriptionQry: string) {`

src/GraphClient.ts
deepesh-kn: Why is our implementation influenced by the unit test utility code?
It should be `throw` instead of `promise.reject`. This is not async function.

test/Subscriber/subscribe.test.ts
```diff
@@ -0,0 +1,40 @@
...
+      sinon.fake.resolves(true),
+    );
+    subscriber = new Subscriber(graphClient, [subscriptionQry]);
+    await subscriber.subscribe();
+
+    assert.strictEqual(
+      subscriber.querySubscriptions.length,
+      1,
+      "Subscription failed!!!"
+    );
```
deepesh-kn: If the `this.graphClient.subscribe` is called with correct params should also be asserted.

test/Subscriber/unsusbcribe.test.ts
```diff
@@ -0,0 +1,51 @@
+'use strict';
+
+const sinon = require('sinon');
+import { assert } from 'chai'
+import { createMockClient } from 'mock-apollo-client';
```
deepesh-kn: Why are we using `mock-apollo-client`? 
How does it help in this test?

test/Subscriber/unsusbcribe.test.ts
```diff
@@ -0,0 +1,51 @@
...
+    apolloClient = createMockClient();
+    graphClient = new GraphClient(apolloClient);
+    subscriptionQry = 'subscription{stakeRequesteds{id}}';
+  });
+
+  it('should work with correct parameters', async () => {
+    sinon.replace(
+      graphClient,
+      'subscribe',
+      sinon.fake.resolves(new SpyUnsubscribe()),
```
deepesh-kn: I think that we should spy on the actual `querySubscription` object, instead of using `SpyUnsubscribe`.

test/Subscriber/unsusbcribe.test.ts
abhayks1: https://www.npmjs.com/package/mock-apollo-client 
Helps unit test components which use the Apollo Client.

src/Subscriber.ts
```diff
@@ -0,0 +1,52 @@
...
+
+  private graphClient: GraphClient;
+
+  /**
+   * Constructor
+   *
+   * @params {GraphClient} graphClient graph client instance.
+   * @param {string[]} subscriptionQueries Array of subscription queries.
+   */
+  public constructor(graphClient: GraphClient, subscriptionQueries: string[]) {
```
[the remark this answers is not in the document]
abhayks1: As per discussion we will make this change based on your POC results.
It's hard to mock/test without dependency injection: https://blog.risingstack.com/dependency-injection-in-node-js/

test/GraphClient/subscribe.test.ts
```diff
@@ -0,0 +1,40 @@
...
+  it('should work with correct parameters', async () => {
+    const spySubscribe = sinon.replace(
+      apolloClient,
+      'subscribe',
+      sinon.fake.returns({
+        subscribe: () => Promise.resolve(true)
+      }),
+    );
+    const querySubscriber = await graphClient.subscribe(subscriptionQry);
+    assert(querySubscriber);
```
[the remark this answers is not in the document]
abhayks1: Done with 2nd point. I had hard time asserting the 1st point 😓 . 
I believe apollo client behaves differently with the method call.  this.apolloClient.next()..next().

A code example will be helpful.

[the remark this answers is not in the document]
abhayks1: Done with first point also. We needed to parse subscription queries with gql(Graph query language)

test/Subscriber/unsusbcribe.test.ts
abhayks1: For that Subscription object needs to be created. Here apollo exposes Subscription as type and not class. I think that class is encapsulated in the apollo client. Any reference for creating Subscription object will be helpful.

deepesh-kn: Please check `sinon.mock` for this.

test/Subscriber/unsusbcribe.test.ts
deepesh-kn: Please use `sinon.mock` for this.

abhayks1: Done!

abhayks1:
Available for review again!
Should be merged after merge of below PR
https://github.com/OpenST/facilitator/pull/65

src/GraphClient.ts
```diff
@@ -0,0 +1,79 @@
...
+export default class GraphClient {
+  private apolloClient: ApolloClient<NormalizedCacheObject>;
+
+  /**
+   * GraphClient constructor. It expects apollo client as input. Apollo Client is a fully-featured,
+   * production ready caching GraphQL client for every UI framework and GraphQL server.
+   *
+   * @param {ApolloClient<NormalizedCacheObject>} apolloClient Apollo client for subscription.
+   */
+  public constructor(apolloClient: ApolloClient<NormalizedCacheObject>) {
```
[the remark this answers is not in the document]
deepesh-kn: Marking resolved, this is not a blocker, we can revisit this later.

PR #73, opened 2019-06-19
StakeRequest model (with a help of sequelize)

pgev:
Implements stake request model classes.

`StakeRequestAttributes`: Represents an interface for an input to create `StakeRequest` objects in `StakeRequestRepository`. 

`StakeRequest`: Represents a stake request object to be used throughout facilitator.

`StakeRequestModel`: Represents a `StakeRequest` database model. Allows to initialize stake request database table's schema and in addition represents a raw in that table.

`StakeRequestRepository`: Allows to create and retrieve stake request object. A class should be an interface for other classes in facilitator to interact with stake request model.

`Database`: Creates, initializes and gives an access to different models throughout facilitator.

Fixes: #53


pgev:
> This looks very clean 👍
> 
> **Questions:**
> 
> 1. Why do we need `StakeRequestAttributes`? Is it that any class that implement the `StakeRequestAttributes` interface can be passed in the `build` and `create` params?

Yes. StakeRequestAttributers is an interface for an input to create StakeRequest object.

> 2. What `StakeRequestRepository.build` does?

`build` pattern in sequelize creates an object, which is not yet saved in database model and can be separately (in some other time) saved into database. However, I've decided to remove that for now. We can add it later if needed.

> 3. What do you think if we do the following?
>    
>    * `StakeRequestRepository` can extends `Model`
>    * `StakeRequestModel` can be a simple class that has all the properties (may be some functions to implement business logic). Implements/confirms to the `StakeRequestAttributes` interface.
>    * `StakeRequestModel` can be passed as params in `StakeRequestRepository.build` and `StakeRequestRepository.create`
>    * `StakeRequestModel` is returned from `StakeRequestRepository.get` function.
> 
> In this way there is clear separation of Model objects(related to facilitator) and Repository (sequelize model objects)

In the current design, StakeRequestModel is a sequelize model, which in my understanding represents two things: a database schema model, its `init` function defines database table schema and database table raw (all `find*` functions returns Model object).

Currently, for abstracting away from underlying database representation there are StakeRequestAttributes, StakeRequest and StakeRequestRepository classes. I've made some changes after your comment, please, have a look and we could discuss this point one more time. Thx!

abhayks1:
Started reviewing 🎸 🚗

package.json
```diff
@@ -31,23 +31,30 @@
...
     "eslint-config-airbnb-base": "13.1.0",
-    "eslint-plugin-import": "2.14.0",
+    "eslint-plugin-import": "2.17.3",
     "eslint-plugin-json": "1.4.0",
     "mocha": "6.1.4",
     "nyc": "14.1.1",
-    "sinon": "7.3.2"
+    "sequelize": "5.8.10",
+    "sinon": "7.3.2",
+    "sqlite3": "4.0.8"
```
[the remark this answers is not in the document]
pgev: Yes, we still need sqlite3 (according to sequelize documentation). However, @types/sqlite3 can be removed.

src/models/Database.ts
```diff
@@ -0,0 +1,69 @@
...
+export default class Database {
+  /* Storage */
+
+  public stakeRequestRepository: StakeRequestRepository;
+
+
+  /* Public Functions */
+
+  /** Creates in memory database. */
+  public static async createInMemory(): Promise<Database> {
```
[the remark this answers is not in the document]
pgev: Agree. The `create` method has replaced above-mentioned methods that gets `storage` argument to specify either a database file path or ':memory:' in case of in memory database.

src/models/Database.ts
```diff
@@ -0,0 +1,69 @@
...
+
+    const db = new Database(sequelize);
+    await sequelize.sync();
+
+    return db;
+  }
+
+  /** Creates a database from a file. */
+  public static async createFromFile(dbFilePath: string): Promise<Database> {
+    const sequelize = new Sequelize({
```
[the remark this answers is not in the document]
pgev: As has been discussed: application creates a database object on top and passes its down.

src/models/StakeRequestRepository.ts
```diff
@@ -0,0 +1,123 @@
...
+  stakerProxy: string;
+  createdAt: Date;
+  updatedAt: Date;
+}
+
+export class StakeRequestRepository {
+  /* Public Functions */
+
+  public constructor(initOptions: InitOptions) {
+    StakeRequestModel.init(
```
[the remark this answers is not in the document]
pgev: Do you think we should do it now or go with this approach and have ticket to improve it later. I am open for both approaches.

package.json
```diff
@@ -31,23 +31,30 @@
...
+    "chai-as-promised": "7.1.1",
     "eslint": "5.16.0",
     "eslint-config-airbnb-base": "13.1.0",
-    "eslint-plugin-import": "2.14.0",
+    "eslint-plugin-import": "2.17.3",
     "eslint-plugin-json": "1.4.0",
     "mocha": "6.1.4",
     "nyc": "14.1.1",
-    "sinon": "7.3.2"
+    "sequelize": "5.8.10",
```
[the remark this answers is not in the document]
pgev: Yes. I've moved sequelize and sqlite3 as a dependency.

abhayks1:
Revieiwng 🚋

tests/models/StakeRequestRepository/create.test.ts
```diff
@@ -0,0 +1,108 @@
...
+
+import chai = require('chai');
+import chaiAsPromised = require('chai-as-promised');
+chai.use(chaiAsPromised);
+const { assert } = chai;
+
+
+describe('StakeRequestRepository::create', (): void => {
+  it('Checks creation of stake request model.', async (): Promise<void> => {
+    const db = await Database.create();
```
abhayks1: Some of the code can be moved to beforeEach as they are repetitive. 
e.g.
```
const db = await Database.create();
const stakeRequestAttributes: StakeRequestAttributes = {
      stakeRequestHash: 'stakeRequestHashA',
      messageHash: 'messageHashA',
      amount: 1,
      beneficiary: 'beneficiaryA',
      gasPrice: 2,
      gasLimit: 3,
      nonce: 4,
      gateway: 'gatewayA',
      stakerProxy: 'stakerProxyA',
    };

```

tests/models/StakeRequestRepository/get.test.ts
```diff
@@ -0,0 +1,92 @@
...
+} from '../../../src/models/StakeRequestRepository';
+import Database from '../../../src/models/Database';
+
+import Util from './util';
+
+import assert = require('assert');
+
+describe('StakeRequestRepository::get', (): void => {
+  it('Checks retrieval of an existing model.', async (): Promise<void> => {
+    const db = await Database.create();
```
abhayks1: Same comment as above on moving common pieces to beforeEach.

src/models/Database.ts
```diff
@@ -0,0 +1,62 @@
...
+  /**
+   * Creates a database object.
+   *
+   * Creates an empty database if a database does not exist.
+   * Opens a database in read/write mode.
+   *
+   * @param storage A database file path or ':memory' in case of in
+   *                memory database.
+   */
+  public static async create(storage = ':memory:'): Promise<Database> {
```
abhayks1: Should we also be adding test for Database.create() method?

tests/models/StakeRequestRepository/util.ts
```diff
@@ -0,0 +1,77 @@
...
+
+    assert.strictEqual(
+      stakeRequest.gateway,
+      stakeRequestAttributes.gateway,
+    );
+
+    assert.strictEqual(
+      stakeRequest.stakerProxy,
+      stakeRequestAttributes.stakerProxy,
+    );
```
deepesh-kn: For all the asserts above,  `reason` text,  the third param is missing.
We have followed this in all tests.

src/models/Database.ts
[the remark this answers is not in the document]
abhayks1: ok

src/models/StakeRequestRepository.ts
```diff
@@ -0,0 +1,123 @@
...
+        modelName: 'stakeRequest',
+        tableName: 'stake_request',
+      },
+    );
+  }
+
+  /** Creates a stake request model in the repository and syncs with database. */
+  public async create(stakeRequest: StakeRequestAttributes): Promise<StakeRequest> {
+    return await StakeRequestModel.create(stakeRequest) as StakeRequest;
+  }
```
deepesh-kn: I think we are missing `update` and `delete`  operations?

src/models/StakeRequestRepository.ts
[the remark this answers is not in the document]
deepesh-kn: I think we should do this if it is not going to take more than 2-3 hours.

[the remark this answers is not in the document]
pgev: @deepesh-kn I can not easily estimate it. I can work 2-3 hours and if it's not ready I would stop.

src/models/StakeRequestRepository.ts
pgev: What is the case that we need update and delete for stake request?

src/models/Database.ts
pgev: I would go with no. This would mean to test sequelize which is not our case. Also, Database is going to be used throughout all tests and we would test its validity there. That's why I've not done any testing for database separately.

tests/models/StakeRequestRepository/util.ts
pgev: I would propose rather to put a reason test for asserts in test (I am speaking about tests here only), that are not obvious from the test context. In this exact case, from the function name it's already obvious what are those assertions. What are your thoughts?

src/models/StakeRequestRepository.ts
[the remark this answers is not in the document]
pgev: https://github.com/OpenST/facilitator/issues/75

src/models/StakeRequestRepository.ts
deepesh-kn: We may not need delete.
We need an update to store `messageHash` while accepting stake request.

tests/models/StakeRequestRepository/util.ts
deepesh-kn: If anything fails in `checkStakeRequestAgainstAttributes`, it will be difficult to find exactly what has failed just by seeing the test logs.
We may know that by seeing the line number, but here we can add the exact reason as we know what we are comparing.

pgev: You are right: a stack trace will show us the exact position where it has failed.

I still do not think, that adding a reason string to something that is already obvious going to contribute something useful. Function says `checkStakeRequestAgainstAttributes` by passing stakeRequest and stakeRequestAttributes and then simply comparing member by member. Adding `Stake request's staker proxy does not match with stake request attributes' staker proxy` is verbose and does not contribute something useful.

src/models/StakeRequestRepository.ts
pgev: Agree. Added updateMessageHash function with tests.

tests/models/StakeRequestRepository/util.ts
deepesh-kn: When I will run all the tests, I will not know if the broken test is related to `checkStakeRequestAgainstAttributes`.  Our practice was to always put the reason.

src/models/Database.ts
deepesh-kn: I think we should have tests for all the functions we write.
I see the point here that it is not straight forward to test this function as the objects are created and it is not easy to mock it.

I will do a POC and try to find out how we can test such functions.

src/models/Database.ts
```diff
@@ -0,0 +1,62 @@
...
+    const db = new Database(sequelize);
+    await sequelize.sync();
+
+    return db;
+  }
+
+
+  /* Private Functions */
+
+  public constructor(sequelize: Sequelize) {
```
deepesh-kn: Documentation is missing
Line 52 says its private function.

src/models/StakeRequestRepository.ts
```diff
@@ -0,0 +1,184 @@
...
+  amount: number;
+  beneficiary: string;
+  gasPrice: number;
+  gasLimit: number;
+  nonce: number;
+  gateway: string;
+  stakerProxy: string;
+}
+
+export interface StakeRequest extends StakeRequestAttributes {
```
deepesh-kn: It is difficult to understand the use of `StakeRequest` and `StakeRequestAttributes`
Documentation for these will be helpful.

src/models/Database.ts
pgev: Sure.

tests/models/StakeRequestRepository/util.ts
pgev: @deepesh-kn, sorry I am not sure I follow you.

I have changed the Utils::checkStakeRequestAgainstAttributes 

```ts
    assert.strictEqual(
      stakeRequest.stakerProxy,
      `${stakeRequestAttributes.stakerProxy}1`,
    );
```

here is the output when I run test:

```bash
- 'stakerProxy'
+ 'stakerProxy1'
      + expected - actual

      -stakerProxy
      +stakerProxy1
      
      at Object.checkStakeRequestAgainstAttributes (tests/models/StakeRequestRepository/util.ts:69:12)
      at Object.<anonymous> (tests/models/StakeRequestRepository/updateMessageHash.test.ts:83:10)
      at Generator.next (<anonymous>)
      at fulfilled (tests/models/StakeRequestRepository/updateMessageHash.test.ts:19:58)
```

or you've meant something else?

src/models/Database.ts
pgev: Ok.

src/models/StakeRequestRepository.ts
pgev: Ok.

tests/models/StakeRequestRepository/util.ts
deepesh-kn: We can take this discussion offline, I think this should not block the PR from merging.

PR #84, opened 2019-06-26
Object creations flow

### LINKED ISSUE #56: Implement transaction handler

sarvesh-ost:
fixes #56 
This PR adds an proposal about where to instantiate services and how to inject them to handler. 

⚠️ The services created in this PR is a dummy services should be replaced with actual services after implementation.

Merged after #68


sarvesh-ost:
No more needed in favor of container approach.

PR #88, opened 2019-06-27
implemented unlock method and added test cases

puneet-khushwani-eth:


test/Account/unlock.test.ts
```diff
@@ -0,0 +1,39 @@
...
+    accountObject = new Account(accountCreationResponse.address, accountCreationResponse.encryptedKeyStore);
+  });
+
+  it('should unlock successfully with valid password', () => {
+    assert.strictEqual(
+      accountObject.unlock(web3, validPassword),
+      true,
+      'should return true for successful unlock',
+    );
+  });
```
deepesh-kn: Please spy and check if the following methods were called.
```
 const web3Account = web3.eth.accounts.decrypt(this.encryptedKeyStore, password);
web3.eth.accounts.wallet.add(web3Account);
```

test/Account/unlock.test.ts
```diff
@@ -0,0 +1,63 @@
+import * as sinon from 'sinon';
+import { assert } from 'chai';
+
+import Account from '../../src/Account';
+import SpyAssert from '../utils/SpyAssert';
+
+const Web3 = require('web3');
+
+describe('Unlock', () => {
+  let web3: any;
```
abhayks1:
```suggestion
  let web3: Web3;
```

test/Account/unlock.test.ts
```diff
@@ -0,0 +1,63 @@
...
+    validPassword = 'validPassword';
+    accountCreationResponse = Account.create(web3, validPassword);
+    accountObject = new Account(
+      accountCreationResponse.address,
+      accountCreationResponse.encryptedKeyStore
+    );
+  });
+
+  it('should unlock successfully with valid password', () => {
+    const dummyAccount = web3.eth.accounts.create();
```
abhayks1:
```suggestion
    const dummyAccount = await web3.eth.accounts.create();
```

src/Account.ts
```diff
@@ -43,9 +43,16 @@ export default class Account {
    * @param password The password required to unlock the account.
    * @returns `true` if its unlocked otherwise false.
    */
-  // public unlock(web3: Web3, password: string): boolean {
-  //   // Unlocking the account and adding it to the local web3 instance so that everything is signed
-  //   // locally when using web3.eth.send
-  //   return false;
-  // }
+  public unlock(web3: Web3, password: string): boolean {
```
abhayks1: Why this function is not integrated / called from anywhere else?

test/Account/unlock.test.ts
puneet-khushwani-eth: We are downgrading version of web to beta 37 which does not have type script support for all types. We are thus removing using Web3 types in our code

src/Account.ts
puneet-khushwani-eth: The calls to this are in a separate PR.

test/Account/unlock.test.ts
puneet-khushwani-eth: This is a sync command as per https://web3js.readthedocs.io/en/1.0/web3-eth-accounts.html.

PR #91, opened 2019-07-01
ContractEntity model

### LINKED ISSUE #85: Create a new table that keeps track of the last processed entity.
The facilitator gets the entities from the graph node.

The subscription to graph node returns overlapping entities. There should be a way to determine the processed entity and get only the unprocessed entity from the graph node. One way to handle is to use the `uts` updated timestamp.

Create a new table that keeps the track of the updated time stamp of each entity. This table will be updated once the entities are received by the transaction handler.

This table will be used to prepare the graph node query based on the updated timestamp.

gulshanvasnani:
PR contains :- 
1. ContractEntityModel which tracks timestamp per entity and contract address.
2. Unit test cases for each method of ContractEntityRepository.

fixes #85


src/models/UTSRepository.ts
```diff
@@ -0,0 +1,160 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+// ----------------------------------------------------------------------------
+
+/* eslint-disable class-methods-use-this */
```
sarvesh-ost: 😱

test/models/ContractEntityRepository/update.test.ts
```diff
@@ -0,0 +1,112 @@
...
+const { assert } = chai;
+
+interface TestConfigInterface {
+  db: Database;
+}
+
+let config: TestConfigInterface;
+let contractEntityAttributes: ContractEntityAttributes;
+describe('ContractEntityRepository::create', (): void => {
+  function assertion(
```
sarvesh-ost: This function is duplicated from `get` test.

test/models/ContractEntityRepository/update.test.ts
```diff
@@ -0,0 +1,112 @@
...
+    );
+
+    const response = await config.db.contractEntityRepository.get(
+      contractEntityAttributes,
+    );
+
+    assertion(response as ContractEntity, contractEntityAttributes.timestamp);
+  });
+
+  it('should fail for non-existing contract address',
```
sarvesh-ost: It's not failing, right? it's not updating if the record doesn't exist.  Its handling gracefully.

src/models/ContractEntityRepository.ts
```diff
@@ -0,0 +1,189 @@
...
+export interface ContractEntity extends ContractEntityAttributes {
+  createdAt: Date;
+  updatedAt: Date;
+}
+
+/**
+ * Entity types of origin and aux chain for which timestamp will be recorded.
+ */
+export enum EntityType {
+  StakeRequested = 'stakerequested',
```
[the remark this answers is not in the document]
gulshanvasnani: Added list of only those which we handle.

src/models/ContractEntityRepository.ts
```diff
@@ -0,0 +1,189 @@
...
+ * Class enables creation, update and retrieval of ContractEntity objects.
+ * On construction it initializes underlying database model.
+ */
+export class ContractEntityRepository {
+  /* Public Functions */
+
+  public constructor(initOptions: InitOptions) {
+    ContractEntityModel.init(
+      {
+        id: {
```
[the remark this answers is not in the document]
gulshanvasnani: As there can be may entities per contract address, so we cant keep only contract address as primary key.
Currently, contract address and entity type acts as composite unique key.

src/models/ContractEntityRepository.ts
```diff
@@ -0,0 +1,189 @@
...
+    return contractEntity;
+  }
+
+  /**
+   * Updates ContractEntity data in database.
+   * @param contractEntityAttributes ContractEntityAttributes object.
+   * @return Number of affected rows.
+   */
+  public async update(contractEntityAttribute: ContractEntityAttributes): Promise<number[]> {
+    return ContractEntityModel.update(contractEntityAttribute, {
```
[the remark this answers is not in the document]
gulshanvasnani: As of now we need to update the timestamp. But using `contractEntityAttribute` as input will help to update more columns in future. Otherwise we would have to create different update methods for each column.

src/models/ContractEntityRepository.ts
```diff
@@ -0,0 +1,189 @@
...
+              EntityType.MintProgressed,
+              EntityType.GatewayProven,
+              EntityType.BountyChangeInitiated,
+              EntityType.BountyChangeConfirmed,
+            ],
+          }),
+          unique: 'UK_contractaddress_entitytype',
+        },
+        timestamp: {
+          type: DataTypes.BIGINT,
```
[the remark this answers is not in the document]
gulshanvasnani: Using `BIGINT` will help to do comparisons easily.

src/models/ContractEntityRepository.ts
```diff
@@ -0,0 +1,189 @@
...
+export class ContractEntityModel extends Model {
+}
+
+/**
+ * To be used for calling any methods which would change states of record(s) in Database.
+ */
+export interface ContractEntityAttributes {
+  contractAddress: string;
+  entityType: string;
+  timestamp: BigNumber;
```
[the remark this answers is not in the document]
gulshanvasnani: Same as above.

src/models/ContractEntityRepository.ts
```diff
@@ -0,0 +1,189 @@
...
+      return Promise.reject(new Error(`Failed to create a ContractEntity: ${JSON.stringify(errorContext)}`));
+    }
+  }
+
+  /**
+   * Fetches ContractEntity data from database.
+   * @param contractEntityAttribute ContractEntityAttributes object.
+   * @returns ContractEntity object containing values which satisfy the `where` condition.
+   */
+  public async get(
```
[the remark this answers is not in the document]
gulshanvasnani: Primary key is contract address and entity type. It is changed to accept only these two parameters.

src/models/ContractEntityRepository.ts
```diff
@@ -0,0 +1,189 @@
...
+
+  /**
+   * Updates ContractEntity data in database.
+   * @param contractEntityAttributes ContractEntityAttributes object.
+   * @return Number of affected rows.
+   */
+  public async update(contractEntityAttribute: ContractEntityAttributes): Promise<number[]> {
+    return ContractEntityModel.update(contractEntityAttribute, {
+      where: {
+        contractAddress: {
```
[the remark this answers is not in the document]
gulshanvasnani: Yes, updation is done on primary key consisting of fields contract address and entity type.

src/models/ContractEntity.ts
```diff
@@ -0,0 +1,69 @@
...
+import BigNumber from 'bignumber.js';
+import Comparable from "../observer/Comparable";
+
+/**
+ * Represents model object.
+ */
+export default class ContractEntity extends Comparable<ContractEntity> {
+  public contractAddress: string;
+
+  public entityType?: string;
```
abhayks1: entityType is a mandatory field here. we don't need to add "?" as optional field?
Same for timestamp.
created_at, updated_at are optional fields during object creating.

src/repositories/ContractEntityRepository.ts
```diff
@@ -0,0 +1,175 @@
...
+
+  public readonly updatedAt!: Date;
+}
+
+/**
+ * Entity types of origin and aux chain for which timestamp will be recorded.
+ */
+export enum EntityType {
+  StakeRequesteds = 'stakerequesteds',
+  StakeIntentDeclareds = 'stakeintentdeclared',
```
abhayks1: values are not synced with keys 😮

src/repositories/ContractEntityRepository.ts
```diff
@@ -0,0 +1,175 @@
...
+   * from the passed stake request object.
+   *
+   * @param contractEntity Contract entity object to update.
+   *
+   * @returns Newly created or updated contract entity object.
+   */
+  public async save(contractEntity: ContractEntity): Promise<ContractEntity | null> {
+    const definedOwnProps: string[] = Utils.getDefinedOwnProps(contractEntity);
+
+    ContractEntityModel.upsert(
```
abhayks1:
```suggestion
    await ContractEntityModel.upsert(
```

src/repositories/ContractEntityRepository.ts
```diff
@@ -0,0 +1,175 @@
...
+   * If a contract entity does not exist, it creates, otherwise updates.
+   *
+   * Function ignores (does not set to null) undefined (optional) fields
+   * from the passed stake request object.
+   *
+   * @param contractEntity Contract entity object to update.
+   *
+   * @returns Newly created or updated contract entity object.
+   */
+  public async save(contractEntity: ContractEntity): Promise<ContractEntity | null> {
```
abhayks1: I don't see use cases where  save function returns null.

src/repositories/ContractEntityRepository.ts
```diff
@@ -0,0 +1,175 @@
...
+  }
+
+  /**
+   * Fetches ContractEntity data from database if found. Otherwise returns null.
+   * @param contractAddress Address of the contract.
+   * @param entityType Type of the entity.
+   * @returns ContractEntity object containing values which satisfy the `where` condition.
+   */
+  public async get(
+    contractAddress?: string,
```
abhayks1: why "?" 🤔 
contractAdress and entityType needs to be mandatory here.

src/models/ContractEntity.ts
```diff
@@ -0,0 +1,69 @@
...
+
+  public timestamp?: BigNumber;
+
```
abhayks1: createdAt and updatedAt attributes are missing.

test/TransactionFetcher/fetch.test.ts
```diff
@@ -69,7 +69,6 @@ describe('TransactionFetcher.fetch()', () => {
     const transactionFetcher = new TransactionFetcher(graphClient);
 
     const response = await transactionFetcher.fetch(subscriptionResponse);
-    console.log('response:', response);
```
abhayks1: Thank you for correcting my mistake 👏

test/repositories/ContractEntityRepository/util.ts
```diff
@@ -0,0 +1,52 @@
+// Copyright 2019 OpenST Ltd.
```
abhayks1: Let's remove this till it's clarified if license is needed or not.
Either we add to all files or add to none.

test/repositories/ContractEntityRepository/util.ts
```diff
@@ -0,0 +1,52 @@
...
+    assert.strictEqual(
+      expectedContractEntity.entityType,
+      responseContractEntity.entityType,
+    );
+
+    assert.strictEqual(
+      expectedContractEntity.contractAddress,
+      responseContractEntity.contractAddress,
+    );
+  }
```
abhayks1: Add assertions for createdAt and updatedAt fields.

test/repositories/ContractEntityRepository/util.ts
gulshanvasnani: Removed from files in the PR.

src/repositories/ContractEntityRepository.ts
gulshanvasnani: Changed it to mandatory.

src/models/ContractEntity.ts
```diff
@@ -0,0 +1,69 @@
...
+  /**
+   * Compares ContractEntity objects.
+   * @param {ContractEntity} other ContractEntity object which is to be compared.
+   * @returns {number} `0` if the objects are same, 1 if new object is greater and -1 if new object is lesser.
+   */
+  public compareTo(other: ContractEntity): number {
+
+    const currentObject = this.contractAddress.concat(this.entityType!);
+
+    const newObject = other.contractAddress.concat(other.entityType!);
```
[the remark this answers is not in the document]
gulshanvasnani: Changed it. Thanx 👍

src/models/ContractEntity.ts
```diff
@@ -0,0 +1,69 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+// ----------------------------------------------------------------------------
+
+/* eslint-disable class-methods-use-this */
```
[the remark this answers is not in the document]
gulshanvasnani: Removed it.

src/models/ContractEntityRepository.ts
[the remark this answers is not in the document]
abhayks1: There is a disadvantage of keeping composite primary key with sequelize.
https://stackoverflow.com/questions/51957860/referencing-a-composite-primary-key-in-a-sequelize-js-seed-model
```
While it is possible to create composite primary keys in Sequelize by specifying primaryKey: true against more than one column (as you have already done above), Sequelize doesn't currently support composite foreign keys, so there is no way to reference a model/table which has composite primary keys.
```

Issue #19, opened 2019-05-29
Facilitator:  Handle SIGTERM


deepesh-kn:
This is an implementation ticket. The actual implementation may vary as we start working on it.
Please check #3  first for any updates in interface and params.

If the process encounters SIGTERM, then it should be intercepted.
- Should inform the child-process about the SIGTERM.
- Wait for child-process to complete
- Exit the process.

PR #92, opened 2019-07-01
Added subscription client callback handlers

### LINKED ISSUE #51: Evaluate how to resubscribe the entities (reconnect on an error)
Evaluate how to resubscribe the entities (reconnect on an error)
- Evaluate if not already covered by NPM module
- Create a followup ticket if required.

sarvesh-ost:
fixes #51  

With the current scope, there is no need to specifically handle different callbacks of subscription handlers.  This PR defines different callbacks and logs it. 
In the future, these handlers can be extended as needed.

### DIFF
--- a/src/GraphClient.ts
+++ b/src/GraphClient.ts
@@ -70,6 +70,8 @@ export default class GraphClient {
       reconnect: true,
     },
     WebSocket);
+
+    GraphClient.attachSubscriptionClientCallbacks(subscriptionClient);
     // Creates WebSocket link.
     const wsLink = new WebSocketLink(subscriptionClient);
     // Instantiate in memory cache object.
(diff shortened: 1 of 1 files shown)


puneet-khushwani-eth:
We might want to think of shutting down the process if the connection couldn't be established after the max number of retries we could make.
Thoughts ?

src/GraphClient.ts
```diff
@@ -79,4 +81,32 @@ export default class GraphClient {
...
+      Logger.info('Connecting to the graph node');
+    });
+    subscriptionClient.onReconnecting(() => {
+      Logger.info('Reconnecting to the graph node');
+    });
+    subscriptionClient.onDisconnected(() => {
+      Logger.info('Disconnected to the graph node');
+    });
+    subscriptionClient.onError((error) => {
+      Logger.error(`Error connecting to graph node. Reason: ${error.message}`);
```
abhayks1: > We might want to think of shutting down the process if the connection couldn't be established after the max number of retries we could make.
> Thoughts ?

We can raise SIGTERM in case of error.  Facilitator will need to stop in this case because max retries has a max limit. We will handle the SIGTERM in the ticket #19

PR #94, opened 2019-07-02
Subscription improvements & Implementation of TransactionFetcher

### LINKED ISSUE #86: Update the subscription logic.
Update the subscription logic.

Currently, the graph node returns the same 100 entities every time the new entity is created.

- Subscribe the graph-node for the entities with the limit as 1, this will be just a trigger point for subscriber that entities are updated.
- Subscriber will read the latest `uts` updated timestamp from the database (see #85), and prepare the query to get the latest entities.
- This should also handle the pagination.

abhayks1:
- Subscribe the graph-node for the entities with the limit as 1, this will be just a trigger point for subscriber that entities are updated.
- Subscriber will read the latest uts updated timestamp from the database (see #85), and prepare the query to get the latest entities.
- Pagination handling is implemented.

Fixes #86


src/Facilitator.ts
```diff
@@ -70,18 +80,38 @@ export default class Facilitator {
...
+  public static getSubscriptionDetails(): SubscriptionInfo {
     return {
       origin: {
-        subGraphEndPoint: 'ws://localhost:8000/subgraphs/name/openst/ost-composer',
+        wsSubGraphEndPoint: 'ws://localhost:8000/subgraphs/name/openst/ost-composer',
+        httpSubGraphEndPoint: 'http://localhost:8000/subgraphs/name/openst/ost-composer',
         subscriptionQueries: {
-          stakeRequested: 'subscription{stakeRequesteds{id amount'
-          + ' beneficiary gasLimit gasPrice gateway nonce staker stakeRequestHash}}',
+          stakeRequested: 'subscription{stakeRequesteds(orderDirection: desc, first: 1){id}}',
```
puneet-khushwani-eth: We can also fetch `contractAddress` from subscription data. 

As the db table we plan to create for keeping track of last processed timestamp for each entity would also have contract address as a column, we should also fetch `contractAddress` from subscription data. Without which we may have problems specially for `stateRootAvailable` (which would with redeem flow implementation have two rows in this table, one for origin and other for aux anchor).  

`stakeRequested: 'subscription{stakeRequesteds(orderDirection: desc, first: 1){id,contractAddress}}',`

src/Facilitator.ts
```diff
@@ -70,18 +80,38 @@ export default class Facilitator {
...
         },
       },
       auxiliary: {
-        subGraphEndPoint: 'ws://localhost:8000/subgraphs/name/openst/ost-composer',
-        subscriptionQueries: { stakeRequested: 'subscription{stakeRequesteds{id}}' },
+        wsSubGraphEndPoint: 'ws://localhost:8000/subgraphs/name/openst/ost-composer',
+        httpSubGraphEndPoint: 'http://localhost:8000/subgraphs/name/openst/ost-composer',
+        subscriptionQueries: {
+          stakeRequested: 'subscription{stakeRequesteds(orderDirection: desc, first: 1){id}}',
```
puneet-khushwani-eth: per above comment similarly

'stakeRequested: 'subscription{stakeRequesteds(orderDirection: desc, first: 1){id, contractAddress}}','

src/Facilitator.ts
```diff
@@ -70,18 +80,38 @@ export default class Facilitator {
...
+        },
+      },
+      fetchQueries: {
+        stakeRequested: 'query ($uts: BigInt!) {\n'
+        + '  stakeRequesteds(where: {uts_gt: $uts}, orderDirection: asc, limit: 100) {\n'
```
puneet-khushwani-eth: In an architecture where we have one facilitator working for only one gateway pair, we should be good with not adding `contractAddress` to these queries.

src/GraphClient.ts
```diff
@@ -46,8 +54,9 @@ export default class GraphClient {
       query: gqlSubscriptionQry,
       variables: {},
     }).subscribe({
-      next(response) {
-        handler.handle(response);
+      async next(response) {
+        const transactions = await fetcher.fetch(response.data);
+        await handler.handle(transactions);
```
puneet-khushwani-eth: Should we add a place holder here to keep track of the place from where we would need to update last processed timestamp in uts table ?

src/TransactionFetcher.ts
```diff
@@ -0,0 +1,35 @@
...
+   *
+   * @param data Data received from subscription.
+   * @return Graph query response from graph node.
+   */
+  public async fetch(data: Record<string, object>): Promise<{data: object}> {
+    const entity = Object.keys(data)[0];
+    const query = Facilitator.getSubscriptionDetails().fetchQueries[entity];
+    // Fetch entity based on uts from uts model and update the variables object
+    // Current dummy value is 0
+    const variables = {
```
puneet-khushwani-eth: I guess we would be adding pagination logic here and would return response only after fetching all batches.

src/Facilitator.ts
abhayks1: Done!

src/Facilitator.ts
abhayks1: Done!

src/Facilitator.ts
abhayks1: Agreed 👍

src/TransactionFetcher.ts
abhayks1: Yes implementation in progress for pagination 📟

src/GraphClient.ts
abhayks1: That place holder will be in individual entity handlers 🤝 .

puneet-khushwani-eth: It might lead to all handlers implementing / calling same piece of code. Can we reconsider having it here ? Similar to us adding the fetch logic here. Thoughts ?

abhayks1: Added comment as placeholder. It needs to be integrated with #91 after it's done

test/TransactionFetcher/fetch.test.ts
```diff
@@ -0,0 +1,56 @@
...
+      contractAddress: '0x0000000000000000000000000000000000000022',
+      uts: 0,
+      limit: 100,
+      skip: 0,
+    };
+    SpyAssert.assert(spyGraphClientQuery, 1, [[query, variables]]);
+
+    sinon.restore();
+  });
+});
```
puneet-khushwani-eth: Not sure how and where to but should we also be adding a test to verify if our pagination logic works fine ?

src/EntityGraphQueries.ts
```diff
@@ -0,0 +1,24 @@
...
+  + '    nonce\n'
+  + '    beneficiary\n'
+  + '    contractAddress\n'
+  + '    blockNumber\n'
+  + '    uts\n'
+  + '  }\n'
+  + '}',
+};
+
+export default EntityGraphQueries;
```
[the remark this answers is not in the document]
abhayks1: All the entity wise queries will be defined in the variable: EntityGraphQueries

src/TransactionFetcher.ts
```diff
@@ -0,0 +1,56 @@
...
+    const entityRecord = data[entity][0];
+    const query = EntityGraphQueries[entity];
+    // Fetch uts based on entity & contract address from ContractEntity model and update the
+    // variables object uts field. <PLACEHOLDER>
+    let skip = 0;
+    let iteration = 0;
+    const transactions: object[] = [];
+    const response: any = {};
+    while (true) {
+      skip *= iteration;
```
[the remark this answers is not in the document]
abhayks1: Good catch 👍

src/Facilitator.ts
```diff
@@ -70,18 +84,30 @@ export default class Facilitator {
    *
    * @return <any> Object containing chain based subscriptionQueries.
    */
-  public static getSubscriptionDetails(): any {
+  public static getSubscriptionDetails(): SubscriptionInfo {
```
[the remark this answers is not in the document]
abhayks1: Added all queries in EntityGraphQueries for fetching since we have the placeholder.
Added remaining subscription queries also.

src/SubscriptionQueries.ts
```diff
@@ -0,0 +1,55 @@
...
+    + '  }'
+    + '}',
+
+    stakeIntentDeclareds: 'subscription{stakeIntentDeclareds(orderDirection: desc, first: 1){'
+    + '   id'
+    + '   contractAddress '
+    + '  }'
+    + '}',
+
+    stateRootAvailables: stateRootAvailableSubscriptionQuery,
```
puneet-khushwani-eth: Shouldn't be added here yet I guess. It might even fail as we haven't defined this entity in schema for origin sub graph yet.

src/EntityGraphQueries.ts
```diff
@@ -19,6 +19,133 @@ const EntityGraphQueries: Record<string, string> = {
...
   + '}',
+
+  stakeIntentDeclareds: 'query ($contractAddress: String!, $uts: BigInt!, $skip: BigInt!, $limit:'
+  + ' BigInt!) {\n'
+  + 'stakeIntentDeclareds(skip: $skip, first:$limit, orderDirection: asc, where: {contractAddress:'
+  + ' $contractAddress, uts_gt: $uts}, '
+  + '    orderDirection: asc,'
+  + '    limit: 100) {\n'
+  + '    id\n'
+  + '    amount\n'
```
puneet-khushwani-eth: If we have a select * kinda thing in the query it would help us here.

test/TransactionFetcher/fetch.test.ts
abhayks1: Done!

src/EntityGraphQueries.ts
abhayks1: ```One of the main benefits of using GraphQL is actually to prevent overfetching.```
https://stackoverflow.com/questions/34199982/how-to-query-all-the-graphql-type-fields-without-writing-a-long-query

src/SubscriptionQueries.ts
abhayks1: Done!

PR #95, opened 2019-07-03
Anchor handler

### LINKED ISSUE #59: Implement an anchor service.
Implement an anchor service (to handle the state root commit).

- nudge anchor service (chain Id)
- this will nudge the gateway service.
- Identifies the messages that were waiting for the state root commit.

sarvesh-ost:
fixes #59 

This PR adds anchor handler which decides if it needs to update latestoriginBlockHeight.


puneet-khushwani-eth:
Starting to review 👀

src/handlers/AnchorHandler.ts
```diff
@@ -0,0 +1,76 @@
...
+   */
+  public async persist(transactions: any[]): Promise<AuxiliaryChain[]> {
+    const chainRecord = await this.auxiliaryChainRepository.get(this.auxiliaryChainID);
+    if (chainRecord == null) {
+      throw new AuxiliaryChainRecordNotFoundException('Cannot find record for auxiliary chain');
+    }
+
+    let anchorBlockHeight = chainRecord.lastOriginBlockHeight;
+    transactions
+      .filter(transaction => chainRecord.anchorAddress === transaction.contractAddress)
```
pgev: Do we already filter this in graph node (one could specify in the graph node's manifest from which contract address events should arrive)?

src/handlers/AnchorHandler.ts
```diff
@@ -0,0 +1,76 @@
...
+          anchorBlockHeight === undefined
+          || anchorBlockHeight.lt(new BigNumber(filteredTransaction._blockHeight))
+        ) {
+          anchorBlockHeight = new BigNumber(filteredTransaction._blockHeight);
+        }
+      });
+
+    // No change in block height of interested anchor.
+    if (anchorBlockHeight === undefined) {
+      return [chainRecord];
```
pgev: I am not sure why don't we return here an empty array? No updates has been made.

src/handlers/AnchorHandler.ts
```diff
@@ -0,0 +1,76 @@
...
+        if (
+          anchorBlockHeight === undefined
+          || anchorBlockHeight.lt(new BigNumber(filteredTransaction._blockHeight))
+        ) {
+          anchorBlockHeight = new BigNumber(filteredTransaction._blockHeight);
+        }
+      });
+
+    // No change in block height of interested anchor.
+    if (anchorBlockHeight === undefined) {
```
pgev: Based on documentation for this check: even if anchorBlockHeight is *not* undefined, it could be that after parsing all passed transaction objects no change appears. If chainRecord.lastOriginBlockHeight was not undefined and all _blockHeight in all parsed transactions is less than chainRecord.lastOriginBlockHeight then no change occurs.

src/handlers/AnchorHandler.ts
```diff
@@ -0,0 +1,76 @@
...
+    }
+
+    const hasChanged = chainRecord.lastOriginBlockHeight === undefined
+      || !chainRecord.lastOriginBlockHeight.eq(anchorBlockHeight);
+
+    if (hasChanged) {
+      this.auxiliaryChainRepository.update(chainRecord);
+    }
+
+    return [chainRecord];
```
pgev: Can we please document in which case we do return [chainRecord] ? In case of update or?

src/handlers/AnchorHandler.ts
```diff
@@ -0,0 +1,76 @@
...
+          anchorBlockHeight = new BigNumber(filteredTransaction._blockHeight);
+        }
+      });
+
+    // No change in block height of interested anchor.
+    if (anchorBlockHeight === undefined) {
+      return [chainRecord];
+    }
+
+    const hasChanged = chainRecord.lastOriginBlockHeight === undefined
```
pgev: For code readability I would propose to have hasChanged defined upper before we parse transactions and set it to true in line 48 where we setup anchorBlockHeigh to the new value (which means, there is a change). What do you think?

src/handlers/HandlerFactory.ts
```diff
@@ -8,11 +9,15 @@ export default class HandlerFactory {
    *
    * @return Different kinds of transaction handlers.
    */
-  public static get(repos: Repositories): {stakeRequesteds: StakeRequestedHandler} {
+  public static get(repos: Repositories): { stakeRequesteds: StakeRequestedHandler; anchor: AnchorHandler } {
```
pgev: I feel inconsistency in naming convention, however, it might be only me :), ignore if that's too subjective to me:

I would propose, either (looks better to me).
   StakeRequestHandler
   AnchorHandler
or
   StakeRequestedHandler
   AnchoredHandler

src/handlers/AnchorHandler.ts
```diff
@@ -0,0 +1,76 @@
...
+      || !chainRecord.lastOriginBlockHeight.eq(anchorBlockHeight);
+
+    if (hasChanged) {
+      this.auxiliaryChainRepository.update(chainRecord);
```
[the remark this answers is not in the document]
pgev: Maybe I am missing something, however, do we modify `chainRecord` after reading it from database?

src/handlers/AnchorHandler.ts
sarvesh-ost: Yes, Look into the latest implementation.

src/handlers/AnchorHandler.ts
sarvesh-ost: Refer latest implementation

src/handlers/AnchorHandler.ts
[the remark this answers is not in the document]
sarvesh-ost: Yes, we are reading it and updating it again.

src/handlers/AnchorHandler.ts
sarvesh-ost: 1. Graph node is an external system. 
2. We have contract Address as a field in thegraph entity.  Multiple anchor contracts events can be written to the same anchor entity.

src/handlers/HandlerFactory.ts
pgev: Why don't we change also stakeRequesteds to stakeRequests?

src/handlers/AnchorHandler.ts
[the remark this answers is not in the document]
pgev: I might still miss something, however, should we have the following before updating chainRecord:

chainRecord.lastOriginBlockHeight = anchorBlockHeight;

src/handlers/HandlerFactory.ts
sarvesh-ost: It's because of the response from the graph node. It returns `stakeRequesteds` for data. Please refer TransactionHandler Class.

pgev: Ah, it has a long tail :). Resolving this conversation.

src/handlers/AnchorHandler.ts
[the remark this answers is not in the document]
sarvesh-ost: Oh yes 😱  🤦‍♂ 
I asserted in my test that it should update the blockHeight. But the test was passing because it was updating the same reference. I fixed the test as well. Thanks.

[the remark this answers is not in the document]
pgev: Thx for update :).

PR #101, opened 2019-07-10
Handle SIGINT

### LINKED ISSUE #41: Facilitator: Handle SIGTERM.
This is an implementation ticket. The actual implementation may vary as we start working on it.
Please check #5 first for any updates in interface and params.

If the process encounters `SIGTERM`, then it should be intercepted.
- This should wait for any ongoing transactions to complete.
- unsubscribe from listening to the graph-node entities.
- close all the open connections (Implementer can think if there are any such open connections. e.g: DB).
- Exit the process.

gulshanvasnani:
PR contains changes to handle SIGINT in facilitator start.

fixes #41


src/bin/facilitator-start.ts
```diff
@@ -4,10 +4,21 @@ import Facilitator from '../Facilitator';
 import { Config } from '../Config';
 import FacilitatorStart from '../OptionParser/FacilitatorStart';
 
-
 const facilitatorCmd = commander
   .arguments('[origin_chain] [aux_chain_id]');
 
+process.on('SIGINT', terminationHandler);
```
sarvesh-ost: Should we handle SIGTERM also if some kills for `kill` command?

src/bin/facilitator-start.ts
```diff
@@ -4,10 +4,21 @@ import Facilitator from '../Facilitator';
...
+
+let facilitator: Facilitator;
+async function terminationHandler() {
```
sarvesh-ost: Add return type.

src/bin/facilitator-start.ts
sarvesh-ost: Linter error: No use before define.

PR #103, opened 2019-07-10
Contract entity integration

### LINKED ISSUE #98: Integrate ContractEntityRepository in TransactionFetcher and GraphClient
After PR #91 is merged, use ContractEntityRepository in TransactionFetcher.
1. In fetch method of TransactionFetcher, use `get` method of ContractEntityRepository to get the timestamp. Placeholder is at https://github.com/OpenST/facilitator/blob/master/src/TransactionFetcher.ts#L30

2. In subscribe method of GraphClient, use `save`method of ContractEntityRepository for updating the timestamp. Placeholder is at https://github.com/OpenST/facilitator/blob/master/src/GraphClient.ts#L60

sarvesh-ost:
fixes #98 

With this PR Transaction fetcher uses latest UTS for query and after handler handles transactions latest UTS is updated to contract entity.


src/GraphClient.ts
```diff
@@ -50,25 +55,68 @@ export default class GraphClient {
...
+    contractEntityRepository: ContractEntityRepository,
+  ): Promise<void> {
+    const savePromises = Object.keys(transactions).map(
+      async (transactionKind) => {
+        const { contractAddress } = subscriptionResponse[transactionKind][0];
+        const transaction = transactions[transactionKind].length > 0
+          ? transactions[transactionKind][transactions[transactionKind].length - 1]
+          : null;
+
+        const number = transaction ? transaction.uts : 0;
```
pgev: I would rather not use `number` as a member variable. Also, transaction !== null looks to me better as a check condition.

```ts
        const currentUTS = new BigNumber(
          transaction !== null ? transaction.uts : 0
        );
```

pgev: Also, fallback to 0 seems "strange" to me (as a UTC value). Is this an error situation?

src/GraphClient.ts
```diff
@@ -50,25 +55,68 @@ export default class GraphClient {
...
+   * @param contractEntityRepository Instance of contract entity repository.
+   */
+  private static async updateLatestUTS(
+    transactions: Record<string, Record<string, any>>,
+    subscriptionResponse: Record<string, any[]>,
+    contractEntityRepository: ContractEntityRepository,
+  ): Promise<void> {
+    const savePromises = Object.keys(transactions).map(
+      async (transactionKind) => {
+        const { contractAddress } = subscriptionResponse[transactionKind][0];
```
pgev: Can it be that subscriptionResponse[transactionKind] is empty? Then indexing 0 value will crash.

src/GraphClient.ts
```diff
@@ -50,25 +55,68 @@ export default class GraphClient {
...
-        Logger.error(err);
-      },
-    }));
+    });
+    const querySubscriber = await Promise.resolve(
+      observable
+        .subscribe({
+          async next(response) {
+            const transactions = await fetcher.fetch(response.data);
+            await handler.handle(transactions);
```
pgev: Does it matter that if handler fails to handle some inputs we still update their UTS - s in ContractEntityRepository?

src/GraphClient.ts
```diff
@@ -50,25 +55,68 @@ export default class GraphClient {
...
-      error(err) {
-        // Log error using logger
-        Logger.error(err);
-      },
-    }));
+    });
+    const querySubscriber = await Promise.resolve(
+      observable
+        .subscribe({
+          async next(response) {
```
pgev: Can we add types to `response`, `transactions` to ease readability?

src/GraphClient.ts
```diff
@@ -50,25 +55,68 @@ export default class GraphClient {
...
+          transactionKind,
+          currentUTS,
+        );
+        return contractEntityRepository.save(
+          contractEntity,
+        );
+      },
+    );
+
+    await Promise.all(savePromises);
```
pgev: Promise.all calls promises in parallel. Do we need to save entries in some order, sequential, or (as those are UTC values)?

We should assure, that we do not update utc to smaller value in ContractEntityRepository. Thoughts?

Another thought, should transaction handler(s) update ContractEntityRepository on successful handle? Thoughts?

src/TransactionFetcher.ts
```diff
@@ -27,15 +33,21 @@ export default class TransactionFetcher {
...
     const query = EntityGraphQueries[entity];
-    // Fetch uts based on entity & contract address from ContractEntity model and update the
-    // variables object uts field. <PLACEHOLDER>
+
+    const contractEntityRecord = await this.contractEntityRepository.get(
+      entityRecord.contractAddress,
+      entity,
+    );
+
+    const uts = contractEntityRecord ? contractEntityRecord.timestamp : 0;
```
pgev: I would rather check against null, and again, the case 0 is possible?

src/GraphClient.ts
sarvesh-ost: I will remove this check because this filed will always be received from `thegraph` node.

src/GraphClient.ts
sarvesh-ost: I guess no Because subscription call back is triggered when there is some data.

src/GraphClient.ts
sarvesh-ost: Yea, we need to discuss this Part. What happens if there is an exception on the handler layer. Do we update the UTS or not. 

@deepesh-kn  is going to create a ticket for this.

src/GraphClient.ts
sarvesh-ost: It will only save for one entity with the highest uts. The query sorts the record from increasing order.

src/TransactionFetcher.ts
sarvesh-ost: Yea, make sense. I am checking and throwing error.

src/GraphClient.ts
sarvesh-ost: Done

src/GraphClient.ts
sarvesh-ost: On L102 selects last transaction of a kind. 
```
transactions[transactionKind][transactions[transactionKind].length - 1]
```

If you see queries in `SubscriptionQueries` class, transactions from an entity are sorted in increasing order of timestamp. If we select the last record, it will always ensure the latest timestamp for that transaction kind.

Secondly, We iterate on different transaction kind and select the last transaction. 
which means `savePromises` will have only one promise for each entity type/transaction kind with the latest timestamp/uts.

sarvesh-ost: Transaction handler only knows about how to handle transactions. This is the responsibility of the subscription module to maintain timestamp /uts. I feel transaction handler should be incognito of this timestamp/uts logic.

PR #108, opened 2019-07-11
Change the type of entitytype field to enum

### LINKED ISSUE #104: In class ContractEntity, change the type of field entitytype from string to enum EntityType
In order to avoid the incorrect value of entity type field, EntityType enum must be used instead of string everywhere in the codebase.

gulshanvasnani:
PR contains changes  :
1. Replace type of entitytype field to EntityType enum.
2. Custom validator to verify entitytype in ContractEntityRepository.
3. Added test-case to fail when invalid entity type is being saved. 

fixes #104


src/models/ContractEntity.ts
```diff
@@ -1,13 +1,14 @@
 import BigNumber from 'bignumber.js';
 import Comparable from '../observer/Comparable';
+import { EntityType } from '../repositories/ContractEntityRepository';
```
pgev: This introduces cyclic dependency: ContractEntity includes ContractEntityRepository and vice versa.
Please, move EntityType enum into ContractEntity.ts, this will eliminate cyclic dependency and will bring type closer to model.

src/repositories/ContractEntityRepository.ts
```diff
@@ -58,18 +58,30 @@ export default class ContractEntityRepository extends Subject<ContractEntity> {
...
               EntityType.StateRootAvailables,
               EntityType.StakeIntentConfirmeds,
               EntityType.StakeProgresseds,
               EntityType.MintProgresseds,
               EntityType.GatewayProvens,
-            ],
-          }),
-          primaryKey: true,
+          ],
+          validate: {
```
pgev: Do we really need this validation? I mean, we've already declared it to be enum and supplied the values for that enumeration. It's like checking the type ... 

I've quickly googled. Please, have a look into this two links:
  - https://github.com/sequelize/sequelize/issues/5765
  - https://github.com/sequelize/sequelize/issues/8965

One proposal is to try to create enum like this:

```ts
type: Sequelize.ENUM(EntityType.StakeIntentDeclareds, ...)
```

And if it's not fixed, we can try this simpler validation mechanism:

```ts
isIn: {
  args: [[ value-goes-here ]],
  msg: 'Invalid entity type is specified.'
}
```

gulshanvasnani: Thanks.
I tried the proposal, but got error `Property 'ENUM' does not exist on type 'typeof Sequelize'.`

Passing `typeValidation: true` to sequelize constructor enables validation of ENUM values.

PR #122, opened 2019-07-12
Container implementation for facilitator 

### LINKED ISSUE #105: Container implementation for facilitator.
Implement the container implementation that will hold the objects at the time of facilitator startup.
This container will hold the following objects

- Web3 objects
- Configs.
- Repositories.
- Subscribers.
- Services.
- Transaction handler

sarvesh-ost:
fixes #105 

This PR instantiate all the dependencies to instantiate facilitator using a container and provide instance of facilitator.


src/Facilitator.ts
```diff
@@ -73,32 +29,10 @@ export default class Facilitator {
    */
   public async stop(): Promise<void> {
     if (this.originSubscriber) {
```
puneet-khushwani-eth: as now `this.originSubscriber` and `this.auxiliarySubscriber` are mandatory properties, we can remove the if condition

src/bin/facilitator-init.ts
```diff
@@ -63,9 +63,11 @@ commander
       } catch (e) {
```
puneet-khushwani-eth: For better readability we can swap code as per below
```suggestion
       if (options.force) {
      FacilitatorConfig.remove(options.chainId);
    } else {
      try {
        if (FacilitatorConfig.isFacilitatorConfigPresent(options.chainId)) {
          Logger.error('facilitator config already present. use -f option to override the existing facilitator config.');
          process.exit(1);
        }
      } catch (e) {
        Logger.info('creating facilitator config as it is not present');
      }
    }
```

puneet-khushwani-eth:
Changes look good. Please resolve the conflicts and we can then merge this.

PR #133, opened 2019-07-18
Facilitator start testing fixes

abhayks1:
PR covers fixes which were encountered while testing 
- Facilitator init 
- Facilitator start

After apply these fixes:
- Facilitator start is working
- Facilitator stop is working after doing control+C

### DIFF
--- a/src/subscriptions/Subscriber.ts
+++ b/src/subscriptions/Subscriber.ts
@@ -3,6 +3,7 @@ import GraphClient from './GraphClient';
 import TransactionHandler from '../TransactionHandler';
 import TransactionFetcher from './TransactionFetcher';
 import ContractEntityRepository from '../repositories/ContractEntityRepository';
+import Logger from '../Logger';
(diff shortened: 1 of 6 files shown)


src/subscriptions/Subscriber.ts
```diff
@@ -47,27 +47,28 @@ export default class Subscriber {
...
         this.fetcher,
         this.contractEntityRepository,
-      );
-    });
+      ).then((querySubscription) => {
+        Logger.info(`Subscription done for entity: ${entity}`);
+        this.querySubscriptions[entity] = querySubscription;
+      }),
+    );
+    await Promise.all(subscriptionPromises);
```
puneet-khushwani-eth: We can try avoiding using `await` in last line of methods and return the Promise instead. This would give caller more power.

abhayks1: Changes done!

PR #161, opened 2019-07-26
Nonce Management

puneet-khushwani-eth:
On analysing we found that .send method we use is not always able to take pending pool in consideration before deciding nonce. 

Proposal with this PR is to implement small logic which would once fetch nonce from chain and persist in-memory and for later calls increase this in-memory variable and not go to chain (similar to what we had in faucet repo).


abhayks1:
Started reviewing...

test/Account/getNonce.test.ts
```diff
@@ -0,0 +1,55 @@
...
+    };
+  });
+
+  afterEach(async (): Promise<void> => {
+    sinon.restore();
+  });
+
+  it('should return nonce by calling getTransactionCount on web3', async (): Promise<void> => {
+    const nonce: BigNumber = await accountObject.getNonce(web3MockObject);
+    assert.notStrictEqual(
```
abhayks1:
```suggestion
    assert.deepEqual(
```

test/Account/getNonce.test.ts
```diff
@@ -0,0 +1,55 @@
...
+    SpyAssert.assert(
+      web3MockObject.eth.getTransactionCount,
+      1,
+      [[accountObject.address, 'pending']],
+    );
+  });
+
+  it('should return incremented nonce from in-memory storage', async (): Promise<void> => {
+    const nonce: BigNumber = await accountObject.getNonce(web3MockObject);
+    assert.notStrictEqual(
```
abhayks1:
```suggestion
    assert.deepEqual(
```

src/Utils.ts
```diff
@@ -23,17 +25,22 @@ const Utils = {
    * This method submits a raw transaction and returns transaction hash.
    * @param tx Raw transaction.
    * @param txOption Transaction options.
+   * @param web3 The web3 instance to be used for fetching nonce.
    */
-  async sendTransaction(tx: any, txOption: any): Promise<string> {
+  async sendTransaction(tx: any, txOption: any, web3: Web3): Promise<string> {
```
abhayks1: Should we also add test for sendTransaction method?
I know it should have been written earlier 😞

test/Account/getNonce.test.ts
```diff
@@ -0,0 +1,55 @@
...
+    );
+    SpyAssert.assert(
+      web3MockObject.eth.getTransactionCount,
+      1,
+      [[accountObject.address, 'pending']],
+    );
+  });
+
+  it('should return incremented nonce from in-memory storage', async (): Promise<void> => {
+    const nonce: BigNumber = await accountObject.getNonce(web3MockObject);
```
abhayks1: I feel  accountObject should go to beforeEach. This will make sure unit tests are atomic. Else 2nd unit unit test is dependent on 1st unit test for persistence of memory variable.

We can call getNonce 2 times in the 2nd unit test.

src/Account.ts
```diff
@@ -1,8 +1,12 @@
 import Web3 from 'web3';
```
abhayks1: Please check linting issue in this file. Also please check linting issue in test files.

test/Account/getNonce.test.ts
puneet-khushwani-eth: done.

src/Account.ts
puneet-khushwani-eth: this file has no linter errors. We had checked all test files with the previous PR.

puneet-khushwani-eth:
@abhayks1 Yes I did test start by facilitating a few stake and mints.

src/Utils.ts
puneet-khushwani-eth: We should :) but would request to add it outside of scope of this PR.

PR #165, opened 2019-08-07
Facilitator testing fixes

### LINKED ISSUE #139: Bugs identified during facilitator testing

abhayks1:
Below fixes were identified while testing on chain 1405 and dost app. 

- Account.ts logging improvement on unlock failure.

- CoAnchor address should be checksum in AnchorHandler.

- Filter by Gateway address in StakeRequesthandler.

- MessageRepository non mandatory columns should be Not Null.

Fixes #139


test/TransactionHandler/handle.test.ts
```diff
@@ -8,6 +8,8 @@ import SpyAssert from '../test_utils/SpyAssert';
 import StubData from '../test_utils/StubData';
 
 describe('TransactionHandler.handle()', (): void => {
+  const originChain = '1234';
+  const gatewayAddress = '0x4e4ea3140f3d4a07e2f054cbabfd1f8038b3b4b0';
```
puneet-khushwani-eth: use dummy addresses pls

src/handlers/StakeRequestHandler.ts
```diff
@@ -45,20 +60,25 @@ export default class StakeRequestHandler extends ContractEntityHandler<StakeRequ
...
-        const { stakeRequestHash } = transaction;
-        const amount = new BigNumber(transaction.amount);
-        const beneficiary = Utils.toChecksumAddress(transaction.beneficiary);
-        const gasPrice = new BigNumber(transaction.gasPrice);
-        const gasLimit = new BigNumber(transaction.gasLimit);
-        const nonce = new BigNumber(transaction.nonce);
-        const gateway = Utils.toChecksumAddress(transaction.gateway);
-        const staker = Utils.toChecksumAddress(transaction.staker);
-        const stakerProxy = Utils.toChecksumAddress(transaction.stakerProxy);
+    const gatewayRecord = await this.gatewayRepository.getByChainGateway(
```
puneet-khushwani-eth: if the intent here to verify this gateway exists in db, we should gracefully return if not found. `gatewayRecord!` might not be as readable. 
otherwise `gatewayAddress` was read from mosaic config and passed to this class, we can directly match `transaction.gateway` against it

test/TransactionHandler/handle.test.ts
abhayks1: Changes Done!

src/handlers/StakeRequestHandler.ts
abhayks1: Changes done!

abhayks1:
Available for review again 🍏

src/handlers/Handlers.ts
```diff
@@ -30,21 +30,27 @@ export default class Handlers {
...
-    gatewayProvens: ProveGatewayHandler;
-    stakeProgresseds: StakeProgressHandler;
-    mintProgresseds: MintProgressHandler;
-    stakeIntentConfirmeds: StakeIntentConfirmHandler;
-  } {
+  public static create(
+    repos: Repositories,
+    auxChainId: number,
+    gatewayAddress: string,
+  ): {
```
deepesh-kn: Low priority, this return type can be an interface.

PR #198, opened 2019-09-26
Model alignment with repositories

### LINKED ISSUE #188: Align model optional/mandatory arguments with repository
- There are model attributes which needs to be mandatory but they are optional.
- Model attributes optional/mandatory should be aligned with repository fields.

Scope of this ticket is to go through all models and repository and align whether they should be optional or mandatory.
e.g. MessageTransferRequest model below attributes should be mandatory
```
public amount?: BigNumber;

  public beneficiary?: string;

  public gasPrice?: BigNumber;

  public gasLimit?: BigNumber;

  public nonce?: BigNumber;

  public gateway?: string;

  public sender?: string;

  public senderProxy?: string;

  public messageHash?: string;
```

abhayks1:
There are model attributes which needs to be mandatory but they are optional.
PR covers model attributes alignment with repository fields.

Below models are aligned:
- AuxiliaryChain
- ContractEntity
- Gateway
- Message
- MessageTransferRequest

Fixes #188

### DIFF
--- a/integration_test/faciltiator_start.test.ts
+++ b/integration_test/faciltiator_start.test.ts
@@ -231,6 +231,7 @@ describe('facilitator start', async (): Promise<void> => {
       new BigNumber(1),
       mosaicConfig.auxiliaryChains[auxChainId].contractAddresses.origin.ostEIP20GatewayAddress,
       stakerAccount.address,
+      'senderProxy'
     );
 
     generatedStakeRequestHash = utils.getStakeRequestHash(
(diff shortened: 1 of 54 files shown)


abhayks1:
Available for review 🚀

integration_test/faciltiator_start.test.ts
```diff
@@ -231,6 +231,7 @@ describe('facilitator start', async (): Promise<void> => {
       new BigNumber(1),
       mosaicConfig.auxiliaryChains[auxChainId].contractAddresses.origin.ostEIP20GatewayAddress,
       stakerAccount.address,
+      'senderProxy'
```
sarvesh-ost: Can you change this value to a dummy address.
```suggestion
      'senderProxy',
```

PR #199, opened 2019-09-26
Replaced upsert logic in repositories

### LINKED ISSUE #158: Revisit upsert logic.
Currently, we are using upsert method of sequilze to save the data in the database.
Upsert does not throw any error in case upsert fails.

Revisit the logic so that facilitator can handle the exceptions (Exception handling is not part of this ticket.)

One of the examples of the proposed solution is below:
```
public async save(stakeRequest: StakeRequest): Promise<StakeRequest> {
    const { stakeRequestHash } = stakeRequest;

    let stakeRequestModel = await StakeRequestModel.findOne({
      where: {
        stakeRequestHash,
      },
    });

    const definedOwnProps: string[] = Utils.getDefinedOwnProps(stakeRequest);

    if (stakeRequestModel === null) {
      stakeRequestModel = await StakeRequestModel.create(
        stakeRequest,
        { fields: definedOwnProps },
      );
    } else {
      stakeRequestModel = await stakeRequestModel.update(
        stakeRequest,
        { fields: definedOwnProps },
      );
    }

    const stakeRequestOutput = this.convertToStakeRequest(stakeRequestModel);

    this.newUpdate(stakeRequestOutput);

    return stakeRequestOutput;
  }
```

gulshanvasnani:
PR contains below changes :
1. For all the repositories, upsert logic in save method is replaced with **find-create** and **find-update** query. 
2. Added unit test cases to verify that valid error message is returned when column values cannot be null,but it passed null/undefined.
3. Corrected unit test case for save method of AuxiliaryChainRepository.

fixes #158


src/repositories/GatewayRepository.ts
```diff
@@ -156,18 +156,35 @@ export default class GatewayRepository extends Subject<Gateway> {
...
-    const updatedGateway = await this.get(
-      gateway.gatewayAddress,
-    );
+    let updatedGateway: Gateway|null;
+    if (gatewayModelObj === null) {
+      updatedGateway = this.convertToGateway(await GatewayModel.create(
+        gateway,
+      ));
+    } else {
+      const definedOwnProps: string[] = Utils.getDefinedOwnProps(gateway);
```
abhayks1: getDefinedOwnProps removes column with `undefined` values. Do we need getDefinedOwnProps method since we have model validations in place now?

test/test_utils/assert.ts
```diff
@@ -16,7 +16,26 @@
 
 import chai = require('chai');
 import chaiAsPromised = require('chai-as-promised');
+import { ValidationErrorItem } from "sequelize/types/lib/errors";
 chai.use(chaiAsPromised);
 const { assert } = chai;
 
+export function assertErrorMessages(errorObject: ValidationErrorItem[], messages: string[]) {
```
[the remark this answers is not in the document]
gulshanvasnani: Done.

src/repositories/GatewayRepository.ts
gulshanvasnani: Yes, we need ** getDefinedOwnProps** method.
It will help in partial update of columns for a row.

test/repositories/AuxiliaryChainRepository/save.test.ts
```diff
@@ -93,10 +94,90 @@ describe('AuxiliaryChainRepository::save', (): void => {
...
+    assert.isRejected(
+      config.repos.auxiliaryChainRepository.save(
+        auxiliaryChain,
+      ),
+      'AuxiliaryChain.ostGatewayAddress cannot be null',
+    );
+  });
+
+
+  it('should fail when parameters are undefined', async (): Promise<void> => {
```
abhayks1:
```suggestion
  it('should fail when multiple parameters are undefined', async (): Promise<void> => {
```

test/repositories/GatewayRepository/save.test.ts
```diff
@@ -106,4 +107,88 @@ describe('GatewayRepository::save', (): void => {
...
+
+    assert.isRejected(
+      config.repos.gatewayRepository.save(
+      gateway,
+    ),
+      'Gateway.remoteGatewayAddress cannot be null',
+    );
+  });
+
+  it('should fail when parameters are undefined', async (): Promise<void> => {
```
abhayks1:
```suggestion
  it('should fail when multiple parameters are undefined', async (): Promise<void> => {
```

test/repositories/MessageTransferRequestRepository/save.test.ts
```diff
@@ -159,4 +166,90 @@ describe('MessageTransferRequestRepository::save', (): void => {
...
+
+    assert.isRejected(
+      config.repos.messageTransferRequestRepository.save(
+      requestInput,
+    ),
+      'MessageTransferRequest.gateway cannot be null',
+    );
+  });
+
+  it('should fail when parameters are undefined', async (): Promise<void> => {
```
abhayks1:
```suggestion
  it('should fail when multiple parameters are undefined', async (): Promise<void> => {
```

PR #264, opened 2019-11-13
Bug fixes while testing facilitator

abhayks1:


src/handlers/stake_and_mint/StakeRequestedHandler.ts
```diff
@@ -109,6 +109,7 @@ export default class StakeRequestedHandler extends ContractEntityHandler<Message
             gateway,
             sender,
             senderProxy,
+            null,
```
puneet-khushwani-eth:
```suggestion
            stakeRequest.messageHash || null,
```

puneet-khushwani-eth: also in redeem

puneet-khushwani-eth (changes requested):

puneet-khushwani-eth (approved):
LGTM

Issue #278, opened 2020-02-17
Directory restructuring


deepesh-kn:
- Restructure repository to support different personas 
- Move everything for m0-facilitator
      - For `src`, `test` and `test_integration`

**Please note: using `git mv` command to move the files**

Proposed structure:
```
    facilitator
        |
        |-----src
        |      |----bin
        |      |
        |      |----common (code shared by both gen0 and gen1)
        |      |      |  
        |      |      |-- subscriber
        |      |      .      .
        |      |      .      .
        |      |        
        |      |----mo-faciltiator
        |      |      |-- models
        |      |      |-- services
        |      |      |-- handlers
        |      |      .      .
        |      |      .      .        
        |      |        
        |      |----m1-faciltiator
        |             |-- models
        |             |-- services
        |             |-- handlers
        |             .      .
        |             .      .        
        |      
        |-----test
        |      |----m0-facilitator
        |      |----m1-facilitator
        |      .     .
        |      
        |-----test_integration
        |      |----m0-facilitator
        |      |----m1-facilitator
        |      .     .
        |-----testnet
        |        |-gen0_manifest
        |           |- m1-facilitator.yaml
        |
        |-----package.json
        |-----index.ts
        .      .
        .      .
    ```

abhayks1:
@deepesh-kn @sarvesh-ost  folder names should have underscore as per JS guideline discussed earlier.
e.g. m0_facilitator
m1_facilitator

PR #309, opened 2020-02-18
Implementation of DepositIntent model and repository

abhayks1:
PR covers:
- Implementation of DepositIntent model
- Implementation of  DepositIntent repository
- Tests


abhayks1:
I have commented `npm run test:package` test suite from travis.yml as it's giving web3 errors.
It will be fixed in ticket #278 
Aim is to avoid duplicate effort as @sarvesh-ost was looking into the issue.

deepesh-kn (approved):

PR #322, opened 2020-02-21
Implements GatewayProven handler

### LINKED ISSUE #297: Implement GatewayProven handler
- Upserts `Gateway` model.

pgev:
Fixes #297


pgev:
The implementation is ready (draft PR is linked to the ticket). Tests are missing.
Moving back as someone else can pick.

src/m1_facilitator/handlers/GatewayProvenHandler.ts
```diff
@@ -0,0 +1,112 @@
...
+   * @pre There should exist a record in the `gatewayRepository` for each record
+   *      matching to `record.contractAddress` (in this case gateway address).
+   *
+   * @post Function updates a record of `gatewayRepository` matching to
+   *       `record.contractAddress` (gateway address) by setting
+   *       `remoteGatewayLastProvenBlockNumber` to the `record.blockNumber`
+   *        if `record.blockNumber` is greater than `remoteGatewayLastProvenBlockNumber`
+   *        of the stored value in the repository.
+   */
+  public async handle(records: any[]): Promise<void> {
```
0xsarvesh: We can remove any with specific type.

src/m1_facilitator/handlers/GatewayProvenHandler.ts
```diff
@@ -0,0 +1,112 @@
...
+      const gatewayProvenEntity = new GatewayProvenEntity(
+        record.contractAddress,
+        record.remoteGateway,
+        record.blockNumber,
+      );
+
+      const gatewayModelRecord: Gateway | null = await this.gatewayRepository.get(
+        gatewayProvenEntity.gatewayAddress,
+      );
+      assert(
```
0xsarvesh: I am not sure we want to assert here: 
1. This condition will occur when events from other gateways will appear. This can happen when subgraph deployment has multiple gateway address or wrong address. This essentially is not issue with facilitator codebase. Graph and subgraphs are external systems to facilitator. 
2. If we consider overall architecture,  with this assert, it will throw an error and stop of this execution cycle(callback from subscription). It will not stop facilitator app.  Which also means this will stop execution of other handlers.
 
I think this should just log and silently ignore these records.

src/m1_facilitator/handlers/GatewayProvenHandler.ts
```diff
@@ -0,0 +1,112 @@
...
+   *       `remoteGatewayLastProvenBlockNumber` to the `record.blockNumber`
+   *        if `record.blockNumber` is greater than `remoteGatewayLastProvenBlockNumber`
+   *        of the stored value in the repository.
+   */
+  public async handle(records: any[]): Promise<void> {
+    const savePromises = records.map(async (record): Promise<void> => {
+      const gatewayProvenEntity = new GatewayProvenEntity(
+        record.contractAddress,
+        record.remoteGateway,
+        record.blockNumber,
```
0xsarvesh: type of blocknumber from the subgraph will be string.

src/m1_facilitator/handlers/GatewayProvenHandler.ts
```diff
@@ -0,0 +1,112 @@
...
+        gatewayModel.remoteGA,
+        gatewayModel.gatewayType,
+        gatewayModel.anchorGA,
+        gatewayProvenEntity.blockNumber,
+      );
+
+      await this.gatewayRepository.save(updatedGatewayModel);
+    });
+
+    await Promise.all(savePromises);
```
0xsarvesh, line 109: what will happen when two Gateway proven entity records appear in handle method? 
I think, it will call save twice which intern will trigger services twice. It should only save once only the higher block height.

0xsarvesh (commented):
Looks nice 👍  
Some questions and feedbacks.

src/m1_facilitator/handlers/GatewayProvenHandler.ts
pgev: I've updated implementation and changed GatewayProvenEntity constructor to accept string (it creates a BigNumber object from the string).

src/m1_facilitator/handlers/GatewayProvenHandler.ts
pgev: That's a great point. The current implementation will fail. I've added mutex to guard that.

src/m1_facilitator/handlers/GatewayProvenHandler.ts
pgev: Do we have a typescript type for this or you are proposing to create an interface for that here?

src/m1_facilitator/handlers/GatewayProvenHandler.ts
pgev: I see you point. I've updated code by logging (as a warning) and leaving function on that point.

src/m1_facilitator/handlers/GatewayProvenHandler.ts
0xsarvesh: Proposing a new interface. 

https://github.com/mosaicdao/facilitator/blob/1c9413af912815c908d4e583042fdf2842460bd2/src/m1_facilitator/handlers/AvailableStateRootsHandler.ts#L21

src/m1_facilitator/handlers/GatewayProvenHandler.ts
0xsarvesh: I was proposing to only update block height(remoteGatewayLastProvenBlockNumber)  and ignore lower block height.

src/m1_facilitator/handlers/GatewayProvenHandler.ts
```diff
@@ -0,0 +1,129 @@
...
+        return;
+      }
+
+      // We can safely cast as a check above is excluding the null case.
+      const gatewayModel: Gateway = gatewayModelRecord;
+
+      // As potentially multiple GatewayProven events for the same Gateway
+      // can appear in handler, we lock 'check should we upsert and upsert itself'
+      // part.
+      const release = await this.mutex.acquire();
```
gulshanvasnani: I think if going by `mutex` approach, then acquiring of `mutex` must be done before fetching the gateway record i.e. at line 83.

src/m1_facilitator/handlers/GatewayProvenHandler.ts
```diff
@@ -0,0 +1,111 @@
...
+  /** Constructs GatewayProvenHandler from the specified arguments. */
+  public constructor(gatewayRepository: GatewayRepository) {
+    this.gatewayRepository = gatewayRepository;
+  }
+
+  /**
+   * handle() function accepts an array subgraph gateway-proven entities.
+   * If there is no model in the gateway repository matching to a gateway
+   * global address of an entity, the function logs a warning without failing.
+   * Otherwise, it updates a model in the gateway repository matchin to a
```
deepesh-kn:
```suggestion
   * Otherwise, it updates a model in the gateway repository matching to a
```

src/m1_facilitator/handlers/GatewayProvenHandler.ts
```diff
@@ -0,0 +1,111 @@
...
+   * gateway global address of an entity with the new block number (of a gateway
+   * proof). The gateway repository will throw an exception if the newly
+   * proposed block number is less than or equal to the stored one.
+   * The function catches the exception and logs a warning without failing.
+   */
+  public async handle(entities: GatewayProvenEntityInterface[]): Promise<void> {
+    const savePromises = entities.map(async (entity): Promise<void> => {
+      const gatewayProven = new GatewayProven(entity);
```
deepesh-kn, line 76: Why is the significance of the `GatewayProven` class?
We already have `GatewayProvenEntityInterface`

src/m1_facilitator/handlers/GatewayProvenHandler.ts
pgev: Done.

src/m1_facilitator/handlers/GatewayProvenHandler.ts
pgev: Addressed (in a different way).

src/m1_facilitator/handlers/GatewayProvenHandler.ts
pgev: Addressed (in a different way).

src/m1_facilitator/handlers/GatewayProvenHandler.ts
pgev: GatewayProvenEntityInterface represents subgraph entity (like, blockNumber type is `string`). However, GatewayProven class does type checks and conversion of that entity.

src/m1_facilitator/handlers/GatewayProvenHandler.ts
pgev: Done.

deepesh-kn:
We should add one more test related to `mutex`, to be sure that it is providing the expected result.

deepesh-kn:
> We should add one more test related to `mutex`, to be sure that it is providing the expected result.

Let us see how this behaves in the system test. Merging this.

deepesh-kn (approved):

PR #352, opened 2020-03-16
Creation of Origin transaction table while facilitator initialization

### LINKED ISSUE #347: Origin transaction table should be created when facilitator repository is initialized.

gulshanvasnani:
PR contains changes for creation of origin and auxiliary tables.
Note: Variable `modelInstance` is made `any` because couldn't determine the type.

fixes #347


src/m1_facilitator/repositories/TransactionRepository.ts
```diff
@@ -56,13 +56,15 @@ class TransactionModel extends Model {
 export default class TransactionRepository extends Subject<Transaction> {
   private mutex: Mutex;
 
+  private modelInstance: any;
```
0xsarvesh: The only thing we want to avoid here is `any`. Please check sequlize doc and necessary types.

deepesh-kn:
@gulshanvasnani, can you provide more detail about the fix? What was the issue and how `modelInstance` is helping to fix it?

gulshanvasnani:
> @gulshanvasnani, can you provide more detail about the fix? What was the issue and how `modelInstance` is helping to fix it?

The issue was table `origin_transactions` was not getting created in the db as it was using same model i.e. `TransactionModel`. I found out that it was getting overridden when `auxiliary_transactions` table was created. 
Instead of using `init` method of model class, I used `define` method of sequelize. It then created both the tables. 
Sequelize's `define` method creates model named as `modelName` and returns an reference to the created model. So in `modelInstance` variable we have stored the reference to the created model.

src/m1_facilitator/repositories/TransactionRepository.ts
```diff
@@ -196,6 +198,7 @@ export default class TransactionRepository extends Subject<Transaction> {
 
   // eslint-disable-next-line class-methods-use-this
   private convertToTransaction(transactionModel: TransactionModel): Transaction {
+    console.log('transactionModel in convertToTransaction : ', transactionModel);
```
abhayks1: Remove the console.log

src/m1_facilitator/repositories/TransactionRepository.ts
gulshanvasnani: @sarvesh-ost , unable to determine the exact type and work around for it.
I think we can create the ticket for it.
Let me know what do you think.

0xsarvesh (dismissed):
LGTM 👍  

If we agree to add model types in a separate PR. 

CC @deepesh-kn

src/m1_facilitator/repositories/TransactionRepository.ts
```diff
@@ -25,7 +25,7 @@ import Utils from '../../common/Utils';
 /**
  * An interface, that represents an each row of TransactionRepository.
  */
-class TransactionModel extends Model {
+class TransactionModel {
```
deepesh-kn, line 28: Why it should not extend Model?

src/m1_facilitator/repositories/TransactionRepository.ts
```diff
@@ -56,13 +56,16 @@ class TransactionModel extends Model {
 export default class TransactionRepository extends Subject<Transaction> {
   private mutex: Mutex;
 
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  private model: any;
```
deepesh-kn: `transactionModel` instead of `model`?

deepesh-kn (commented):
Few questions.

src/m1_facilitator/repositories/TransactionRepository.ts
0xsarvesh, line 28: We can proably remove this class?

gulshanvasnani: yes, we can remove it.
I would suggest to make `TransactionModelInterface` otherwise we would have to use `any` in `convertToTransaction` transaction method of `TransactionRepository`

src/m1_facilitator/repositories/TransactionRepository.ts
gulshanvasnani: Done

src/m1_facilitator/repositories/TransactionRepository.ts
gulshanvasnani: Now, as we do no initialize/create `TransactionModel` with using `init`, we do not require to extend `Model` class.
Now, for creating model, we use `sequelize.define`.
Also, for creation and updation of the fields, we use `modelInstance`.

0xsarvesh (approved):
LGTM

PR #376, opened 2020-03-22
Added transaction status in TransactionRepository

### LINKED ISSUE #374: Facilitator should not send (retry to send) the transactions if the estimate gas fails.
If a transaction fails, there is no logic to stop the retry. The facilitator will keep retrying it. This can be critical when there will be competing facilitators. The fix can be we can add a status or retry column in the transaction table.

abhayks1:
Fixes #374


src/m1_facilitator/models/Transaction.ts
```diff
@@ -15,22 +15,34 @@
 import BigNumber from 'bignumber.js';
 import Comparable from '../../common/observer/Comparable';
 
+/**
+ * Type of statuses.
+ */
+export enum Status {
```
deepesh-kn: TransactionStatus

src/m1_facilitator/models/Transaction.ts
```diff
@@ -15,22 +15,34 @@
...
+  Pending = 'pending',
+  Success = 'success',
```
deepesh-kn:
```suggestion
Sent = 'sent',
```

src/m1_facilitator/models/Transaction.ts
```diff
@@ -15,22 +15,34 @@
 import BigNumber from 'bignumber.js';
 import Comparable from '../../common/observer/Comparable';
 
+/**
+ * Type of statuses.
```
deepesh-kn: Please add document, when these status are updated.
What is pending, sent, and failed

src/m1_facilitator/models/Transaction.ts
```diff
@@ -15,22 +15,34 @@
...
-  /** Raw transaction object */
+  /** Raw transaction object. */
   public readonly encodedData: string;
 
-  /** Gas price value at which transaction was sent */
+  /** Gas price value at which transaction was sent. */
   public readonly gasPrice: BigNumber;
 
+  /** Transaction status. */
+  public status: Status;
```
deepesh-kn:
```suggestion
  public transactionStatus: TransactionStatus;
```

deepesh-kn (changes requested):

src/m1_facilitator/models/Transaction.ts
abhayks1: Done!

src/m1_facilitator/models/Transaction.ts
abhayks1: Done!

src/m1_facilitator/models/Transaction.ts
abhayks1: Done!

src/m1_facilitator/models/Transaction.ts
abhayks1: Done!

deepesh-kn (approved):

